PR #236, opened 2021-11-06
Feature: data pipelines

donrestarone:
@Pralish here is the feature branch for the data pipeline project


donrestarone:
@Pralish from a functional perspective, you will need to update this PR with the ability for existing CTA's to be edited with the new attributes/logic

Screenshot from 2021-11-06 12-23-52

donrestarone:
@Pralish 
For Array data type, the user should be able to define if its a single choice (select 1 choice from a list of options) or multiple choices. Right now it defaults to multiple choices

donrestarone:
@Pralish 
For the default value on the field (if there is a value to be prepopulated) you should give the user the choice to define their own default value or leave it blank. 

By default, the field should be blank-- but with the value that was entered in the JSON template set as the placeholder value on the field

donrestarone:
@Pralish for input fields, you currently have a few type validations out of the box (string, number etc). Add string + numbers and REGEX pattern as an option to this. 

When the user chooses REGEX pattern, it should render the pattern field along with a link to [Rubular](https://rubular.com/)

donrestarone:
@Pralish There should be no option to 'destroy' the form. This is toggled by the has form attribute.

donrestarone:
@Pralish When the user edits an api_resource-- they should see the file that was already attached when the user submitted the form. Right now it shows a blank state
Screenshot from 2021-11-06 12-41-24

donrestarone:
@Pralish in the apinamespace show view, you should give the user the ability to sort the api resources based on the created_at/updated_at times (you will need to add a columns for this and make it sortable)

donrestarone:
@Pralish when I added a JSON field to the API namespace, this happened: 

I added a key named `foo` that points at an empty object
Screenshot from 2021-11-06 12-47-55


Form renders:
Screenshot from 2021-11-06 12-47-15
 
But when I added some key/value pairs to the empty object, it did not show up as a JSON field in the show view 

Screenshot from 2021-11-06 12-47-13

donrestarone:
@Pralish Great job on getting the dynamic form fields working! That was no small feat. 

Here is my feedback:

### Handle Existing CTA's (before the upgrade)
from a functional perspective, you will need to update this PR with the ability for existing CTA's to be edited with the new attributes/logic

Screenshot from 2021-11-06 12-23-52

### Improve array data type
For Array data type, the user should be able to define if its a single choice (select 1 choice from a list of options) or multiple choices. Right now it defaults to multiple choices

### Flexible prepopulated values
For the default value on the field (if there is a value to be prepopulated) you should give the user the choice to define their own default value or leave it blank. 

By default, the field should be blank-- but with the value that was entered in the JSON template set as the placeholder value on the field

### Improved type validations on string fields
for input fields, you currently have a few type validations out of the box (string, number etc). Add string + numbers and REGEX pattern as an option to this. 

When the user chooses REGEX pattern, it should render the pattern field along with a link to [Rubular](https://rubular.com/)

### Improved meta attributes on string fields
If the user picks numbers only for the input field, it should add the `tel` property to the field so that on mobile the user is given a numpad

### Forms are toggled only
There should be no option to 'destroy' the form. This is toggled by the has form attribute. 

### Ability to see attached files in edit/show views

When the user edits an api_resource-- they should see the file that was already attached when the user submitted the form. Right now it shows a blank state
Screenshot from 2021-11-06 12-41-24


### Sortable / searchable API resources
in the api_namespace show view, you should give the user the ability to sort the api resources based on the created_at/updated_at times (you will need to add a columns for this and make it sortable) and search for API resources. 

### Bug fixes

#### JSON response not showing

when I added a JSON field to the API namespace, this happened: 

I added a key named `foo` that points at an empty object
Screenshot from 2021-11-06 12-47-55


Form renders:
Screenshot from 2021-11-06 12-47-15
 
But when I added some key/value pairs to the empty object, it did not show up as a JSON field in the show view 

Screenshot from 2021-11-06 12-47-13

Pralish:
@donrestarone Thanks. Could you help me clear out few confusions:

> Flexible prepopulated values

Are the values on the JSONeditor placeholders and not the default values? And for arrays, are those values options?

> JSON response not showing

All the primitive properties are getting displayed under properties. Since Object is also one of them, it is getting shown as part of the property. Do you want me to display them separately like the file types? 
 
image

> Add string + numbers and REGEX pattern as an option to this.

The 'text' type will allow numbers as well. It is the default input type if no type is set. 
 
image

The REGEX field is already available as Pattern. I will change the label and add the link to [rubular](https://rubular.com/).   Should I make them available only for type 'text'?

image

> We should always specify params in Violet Rails core features. The CMS allows all params, however.

Do you mean we do need to use strong parameters for the resource controller?

donrestarone:
Hey @Pralish 

1. Yes the JSONeditor placeholders are the default values. But the user should be able to specify at the `form` level if those values should be prepopulated in the fields or not. For example, in for the `people` `api_namespace` we want the first/last names to be blank with placeholders (these are set to the default values) with the dinner_time field prepopulated to 7PM 
2. Yes, this should be separeted out. Primitive values are part of the json and the JSON field (object) should be presented separately. 
3. Cool, in that case you should rename the validation to free text
4. No, REGEX type validation should only be available for that special field that lets you define a regex to test against. REGEX pattern should be shown only when the user selects this choice
5. Yup, always use strong parameters

Pralish:
@donrestarone , Here's the PR for the requested changes

donrestarone:
@Pralish here's my feedback: 

1. Array field needs placeholder (right now its empty) 
2. Array field does not prepopulate. The user should be able to select the prepopulated value (from the possible values)
3. Array field: when the user selects multi/single the change is reflected in the form but not in the form backend. The value here does not change as expected. 
4. Number field prepopulate option is broken. It shows prepopulated values by default and does not allow toggle
5. boolean field allows form submission without checkbox being checked (on the backend I have it as a required field)
6. boolean field does not follow prepopulation rule. When I uncheck the prepopulated option, the checkbox is checked by default
7. file/rich text content should be editable under `Editing api_form`. The user should be able to mark them as required fields with the option to prepopulate. 
8. Remove namespace type option (it's set to a default value) when creating/editing an api namespace
9. Clarify how JSON API compliance is implemented (we will need to make it compliant for reading/creating resources) in the past I have used https://github.com/jsonapi-serializer/jsonapi-serializer

donrestarone:
@Pralish take a look at the [brakeman analysis](https://github.com/restarone/violet_rails/pull/236/checks?check_run_id=4203354930)

I dismissed the alerts for dynamic method execution because the user-defined params are prefixed. Can you add a test that fails if someone removes this prefix? I want to make sure that something destructive like `User.destroy_all` or `Subdomain.destroy_all` cant be called and every method call will be restricted to `*_api_actions`

donrestarone:
@Pralish Here's the feedback after actions were implemented

1. Prepopulate options for array data type are broken. When I toggle between multi/single select the option is saved on the backend but the option I chose is not reflected in the form. Prepopulated value for single select does not work as well. When required is selected, the user is able to submit the form without actually choosing a choice (same issue if a placeholder is chosen, the placeholder is considered a valid choice)

2. The search form for listing API resources should be paginated and the search fields should show the properties of the API namespace (search by property)  Screenshot from 2021-11-14 08-03-06

3. for boolean/checkbox: if the user checks the box the parameter is submitted as "on" this should be "True" and if the user leaves it unchecked, it should be "False". Right now if its unchecked the parameter is not listed Screenshot from 2021-11-14 08-06-09
4. API actions should be listed on its own page/route. The list should be paginated and allow for sorting and should have columns for timestamps, status, and message. The search form should allow for filtering based on the Trigger Event and action type as well
5. For web requests the user should be able to define custom headers (as a JSON object). You can leave the bearer token field in place and let the user include it in the custom headers by accessing it through the variable `SECRET_BEARER_TOKEN` 
6. The subject for the email message should be: `#{trigger_event} #{api_namespace} v#{api_version}`
7. Discard failed api actions / rerun failed API actions should ask for confirmation before execution (a javascript popup)
8. I only tested email api actions, and I noticed that when I submitted the form multiple times I was only seeing 1 API action in the list of API actions run for that API namespace. I recieved the emails but the actions should reflect exactly how many times triggers were fired.

donrestarone:
@Pralish I tried calculating test coverage on this feature by running the `./run_tests.sh` script but its failing locally. See below:
Screenshot from 2021-11-14 08-41-45
 

That said, we should fix the test suite and then calculate test coverage. Please note that we have an SLA with Restarone clients that ensure that additional feature development doesn't lower the test coverage on `master`. So the test coverage goal for this project is 73.23% or greater: 
Screenshot from 2021-11-14 08-48-34

donrestarone:
@Pralish we are 2/3rds of the way done! Here's the latest feedback after fixing the test suite and implementing feedback from before:

1. When I add a file upload + rich text to a form, I get the following crash: Screenshot from 2021-11-15 06-05-52
Screenshot from 2021-11-15 06-05-23
 
2. user defined JSON properties are being rendered as strings in the JSONeditor. These should be parsed and editable like the other fields Screenshot from 2021-11-15 06-09-17

3. Since API actions reference an an `api_resource_id` you should include an ID column  in the HTML table for API resources
4. I was looking at an initialized API action and I got the following crash when I clicked on the resource id: Screenshot from 2021-11-15 06-14-51  Screenshot from 2021-11-15 06-14-54

Pralish:
@donrestarone , Here's the PR for the latest feedback. https://github.com/restarone/violet_rails/pull/248/

> When I add a file upload + rich text to a form, I get the following crash:
This happened because the file field label was empty. I have added a validation for that.

> Dynamically accessing properties in the web request breaks if the property names are capitalized. This should work with lower case or uppercase characters.
I was not able to replicate this. What was the echoed response for this request?

donrestarone:
@Pralish We are almost there! Here's my feedback on this iteration: 

1. There is some flakyness in the redirect action. Sometimes the action is updated to `completed` but in others it is stuck in `initializing`
Screenshot from 2021-11-15 22-17-32
2. For array data type, if the defined array is empty: it should present the user with a free text multi select. Like the one below, but with the option to enter as many free text choices (separated by pressing enter/return). 
3. API namespace -> Move form preview into a bootstrap expand/collapse area
4. API namespace -> Move info area into a bootstrap expand/collapse area
5. When a user attaches a file, they should see a preview (for images/video)
6. Ability to delete API resources/namespaces from their show pages 
7. Move the create new api resource button to api resource list header (match how it looks on api actions -> action workflow)

I think we should start alpha testing this on staging, are you comfortable with committing to these database migrations? If so, we should spec out the upgrade path from CTA's (and also cleaning that up)

app/controllers/concerns/api_actionable.rb
```diff
@@ -0,0 +1,84 @@
...
+
+    serve_file_action.update(lifecycle_stage: 'executing')
+    file_id = helpers.file_id_from_snippet(serve_file_action.file_snippet)
+    file = Comfy::Cms::File.find(file_id)
+    if params[:action] == 'show' && @redirect_action.nil?
+      flash.now[:file_url] = rails_blob_url(file.attachment)
+    else
+      flash[:file_url] = rails_blob_url(file.attachment)
+    end
+    serve_file_action.update(lifecycle_stage: 'complete', lifecycle_message: file.label)
```
donrestarone, line 28: passing a string like `label: #{file.label}  id: #{file.id} mime_type: #{file.mime_type}`  would be better here. Don't quote me on the mime type part of the snippet :laughing:

app/controllers/api/resource_controller.rb
```diff
@@ -30,19 +30,21 @@ def show
...
-    payload = params[:data]
-    api_resource = @api_namespace.api_resources.new(properties: payload)
+    api_resource = @api_namespace.api_resources.new(
+      properties: resource_params[:data]
+    )
     if api_resource.save
-      render json: { code: 200, status: 'OK', object: serialize_resource(api_resource) }
-    else
-      render json: { code: 400, status: api_resource.errors.full_messages.to_sentence }
+        render json: { code: 200, status: 'OK', object: serialize_resource(api_resource) }
```
donrestarone, line 37: @Pralish can we replace the calls to `serialize_resource` with https://github.com/jsonapi-serializer/jsonapi-serializer ?

app/controllers/concerns/api_actionable.rb
```diff
@@ -0,0 +1,84 @@
+module ApiActionable
```
donrestarone, line 1: @Pralish very well done concern!

app/controllers/concerns/api_actionable.rb
```diff
@@ -0,0 +1,84 @@
...
+  def initialize_api_actions
+    clone_actions(api_action_name)
+  end
+
+  def api_action_name
+    return "#{params[:action]}_api_actions".to_sym if ['new', 'update', 'show', 'create', 'destroy'].include?(params[:action])
+  end
+
+  def clone_actions(action_name)
+    @api_namespace.send(action_name).each do |action|
```
donrestarone, line 80: @Pralish Im unsure of what's happening here. Can you unit test methods that call `clone_actions`?

app/helpers/application_helper.rb
```diff
@@ -11,5 +11,16 @@ def page_entries_info(collection)
...
+
+  def execute_actions(resource, class_name)
+    api_actions = resource.send(class_name)
+    api_actions.each do |api_action|
+      api_action.execute_action unless api_action.serve_file? || api_action.redirect?
+    end
+  end
+
+  def file_id_from_snippet(file_snippet)
+    ComfortableMexicanSofa::Content::Renderer.new(:page).tokenize(file_snippet).last[:tag_params]
```
donrestarone, line 24: @Pralish  I wonder where you had to dumpster dive to find this trick

app/models/api_namespace.rb
```diff
@@ -1,11 +1,78 @@
...
+  def redirect_actions(trigger)
+    api_actions.where(action_type: 'redirect', trigger: trigger)
+  end
+
+  def rerun_api_actions
+    executed_api_actions.where(lifecycle_stage: 'failed').each(&:execute_action)
+  end
+
+  def discard_failed_actions
+    executed_api_actions.where(lifecycle_stage: 'failed').update_all(lifecycle_stage: 'discarded')
```
donrestarone, line 76: @Pralish we should add a maintenance action to `lib/tasks/maintenance.rake` to purge these periodically. I'd say once every year as a default.

app/models/create_api_action.rb
```diff
@@ -0,0 +1,3 @@
+class CreateApiAction < ApiAction
+
```
donrestarone, line 2: @Pralish are these classes: `CreateApiAction` `DestroyApiAction` and `ErrorApiAction` being used?

app/models/show_api_action.rb
```diff
@@ -0,0 +1,2 @@
+class ShowApiAction < ApiAction
+end
```
donrestarone, line 2: @Pralish same here for ShowApiAction and UpdateApiAction -- are they being used?

app/views/shared/_file_download.haml
```diff
@@ -0,0 +1,8 @@
+
+- if flash[:file_url]
+    :javascript
+        var link = document.createElement("a");
+        link.target = '_blank'
+        link.download = '';
+        link.href = "#{flash[:file_url]}";
+        link.click();
```
donrestarone, line 8: @Pralish Nice trick! I wonder if this works in all the major browsers.

lib/tasks/api_action.rake
```diff
@@ -0,0 +1,11 @@
+namespace :api_action do
+  desc "subdomain analytics and report tasks"
+  
+  task :rerun_failed_actions => [:environment] do 
+    p "starting rerun_failed_actions task  @ #{Time.now}"
```
donrestarone, line 5: @Pralish we learned a valuable lesson about logging in the last feature release. Nice!

app/helpers/api_forms_helper.rb
```diff
@@ -0,0 +1,86 @@
+module ApiFormsHelper
+  def render_form(id)
+    # usage in cms  {{ cms:helper render_form, 1 }} here 1 is the id
```
donrestarone, line 3: @Pralish has this been unit tested?

test/controllers/resource_controller_test.rb
```diff
@@ -0,0 +1,30 @@
+require "test_helper"
+
+class ResourceControllerTest < ActionDispatch::IntegrationTest
```
donrestarone, line 3: @Pralish can we add a test case that simulates a more complex form submission? Like including a file upload and rich text blob

app/controllers/concerns/api_actionable.rb
Pralish: Thanks @donrestarone

app/models/create_api_action.rb
Pralish:
```
  has_many :create_api_actions, dependent: :destroy
  accepts_nested_attributes_for :create_api_actions, allow_destroy: true

  has_many :show_api_actions, dependent: :destroy
  accepts_nested_attributes_for :show_api_actions, allow_destroy: true

  has_many :update_api_actions, dependent: :destroy
  accepts_nested_attributes_for :update_api_actions, allow_destroy: true

  has_many :destroy_api_actions, dependent: :destroy
  accepts_nested_attributes_for :destroy_api_actions, allow_destroy: true

  has_many :error_api_actions, dependent: :destroy
  accepts_nested_attributes_for :error_api_actions, allow_destroy: true
  ```
  
  @donrestarone , we do use them.  These classes inherit from ApiAction

File: app/helpers/application_helper.rb (hunk 5 above)
Comment (Pralish) [reply app/helpers/application_helper.rb:24]: I had to go through the cms spec. It took me a while but it was worth it. laughing

File: app/controllers/concerns/api_actionable.rb (hunk 4 above)
Comment (Pralish) [reply app/controllers/concerns/api_actionable.rb:80]: @donrestarone , We are using api_namespace's api_actions as a template for the actions of api_resources. This method creates those api_resources' api_actions. This method will get called for [:show, :update, :destroy] actions

**Comment by Pralish** [comment]
@donrestarone, Please review this OTS when you have time:
This will create api_namespaces, api_forms, api_resources fro existing ctas, cta_responses
```
      CallToAction.where(cta_type: 'contact-us').each do |cta|
        p 'Creating api namespace'
    
        params = {
            name: 'inquiries',
            requires_authentication: false,
            version: 1,
            properties: {
                "name": cta.name_placeholder,
                "email": cta.email_placeholder,
                "phone_number": cta.phone_placeholder,
                "message": cta.message_placeholder
            }.to_json
        }
    
        p 'Creating api form'
        api_namespace = ApiNamespace.create(params)
        api_form_params =  { 
            properties: { 
                "name": {
                    "label": cta.name_label,
                    "pattern": "",
                    "required": "0",
                    "field_type": "input",
                    "prepopulate": "0",
                    "type_validation": "free text"
                },
                "email": {
                    "label": cta.email_label,
                    "pattern": "",
                    "required": "1",
                    "field_type": "input",
                    "prepopulate": "0",
                    "type_validation": "email"
                 },
                "message": {
                    "label": cta.message_label,
                    "pattern": "",
                    "required": "1",
                    "field_type": "textarea",
                    "prepopulate": "0",
                    "type_validation": "free text"
                },
                "phone_number": {
                    "label": cta.phone_label,
                    "pattern": "",
                    "required": "0",
                    "field_type": "input",
                    "prepopulate": "0",
                    "type_validation": "free text"
                }
            },
            api_namespace_id: api_namespace.id,
            success_message: cta.success_message,
            failure_message: cta.failure_message,
            submit_button_label: cta.submit_button_label,
            title: cta.title,
            show_recaptcha: true,
        }
    
        ApiForm.create(api_form_params)
    
        p 'creating api resources'
    
        cta.call_to_action_responses.each do |cta_response|
           cta_response_params = {
                    api_namespace_id: api_namespace.id,
                    properties: cta_response.properties,
                    created_at: cta_response.created_at,
                    updated_at: cta_response.updated_at
                }
    
           api_resource =  ApiResource.create(cta_response_params)
        end
      end
    
      CallToAction.where(cta_type: 'collect-email').each do |cta|
        p 'Creating api namespace'
    
        params = {
            name: 'emails',
            requires_authentication: false,
            version: 1,
            properties: {
                "email": cta.email_placeholder
            }.to_json
        }
    
        p 'Creating api form'
        api_namespace = ApiNamespace.create(params)
        api_form_params =  { 
            properties: { 
                "email": {
                    "label": cta.email_label,
                    "pattern": "",
                    "required": "1",
                    "field_type": "input",
                    "prepopulate": "0",
                    "type_validation": "email"
                 }
            },
            api_namespace_id: api_namespace.id,
            success_message: cta.success_message,
            failure_message: cta.failure_message,
            submit_button_label: cta.submit_button_label,
            title: cta.title,
            show_recaptcha: true,
        }
    
        ApiForm.create(api_form_params)
    
        p 'creating api resources'
    
        cta.call_to_action_responses.each do |cta_response|
           cta_response_params = {
                    api_namespace_id: api_namespace.id,
                    properties: cta_response.properties,
                    created_at: cta_response.created_at,
                    updated_at: cta_response.updated_at
                }
    
           api_resource =  ApiResource.create(cta_response_params)
        end
      end
   ```

donrestarone:
@Pralish There's an asset compilation issue blocking the staging deployment, I havnt looked deep into it yet since I'm still reviewing the code. In the meantime, since you have more context on the js stack for this feature-- can you take a look? 
https://github.com/restarone/violet_rails/issues/252

donrestarone:
> @donrestarone, Please review this OTS when you have time: This will create api_namespaces, api_forms, api_resources from existing ctas, cta_responses
> 
> ```
>       CallToAction.where(cta_type: 'contact-us').each do |cta|
>         p 'Creating api namespace'
>     
>         params = {
>             name: 'inquiries',
>             requires_authentication: false,
>             version: 1,
>             properties: {
>                 "name": cta.name_placeholder,
>                 "email": cta.email_placeholder,
>                 "phone_number": cta.phone_placeholder,
>                 "message": cta.message_placeholder
>             }.to_json
>         }
>     
>         p 'Creating api form'
>         api_namespace = ApiNamespace.create(params)
>         api_form_params =  { 
>             properties: { 
>                 "name": {
>                     "label": cta.name_label,
>                     "pattern": "",
>                     "required": "0",
>                     "field_type": "input",
>                     "prepopulate": "0",
>                     "type_validation": "free text"
>                 },
>                 "email": {
>                     "label": cta.email_label,
>                     "pattern": "",
>                     "required": "1",
>                     "field_type": "input",
>                     "prepopulate": "0",
>                     "type_validation": "email"
>                  },
>                 "message": {
>                     "label": cta.message_label,
>                     "pattern": "",
>                     "required": "1",
>                     "field_type": "textarea",
>                     "prepopulate": "0",
>                     "type_validation": "free text"
>                 },
>                 "phone_number": {
>                     "label": cta.phone_label,
>                     "pattern": "",
>                     "required": "0",
>                     "field_type": "input",
>                     "prepopulate": "0",
>                     "type_validation": "free text"
>                 }
>             },
>             api_namespace_id: api_namespace.id,
>             success_message: cta.success_message,
>             failure_message: cta.failure_message,
>             submit_button_label: cta.submit_button_label,
>             title: cta.title,
>             show_recaptcha: true,
>         }
>     
>         ApiForm.create(api_form_params)
>     
>         p 'creating api resources'
>     
>         cta.call_to_action_responses.each do |cta_response|
>            cta_response_params = {
>                     api_namespace_id: api_namespace.id,
>                     properties: cta_response.properties,
>                     created_at: cta_response.created_at,
>                     updated_at: cta_response.updated_at
>                 }
>     
>            api_resource =  ApiResource.create(cta_response_params)
>         end
>       end
>     
>       CallToAction.where(cta_type: 'collect-email').each do |cta|
>         p 'Creating api namespace'
>     
>         params = {
>             name: 'emails',
>             requires_authentication: false,
>             version: 1,
>             properties: {
>                 "email": cta.email_placeholder
>             }.to_json
>         }
>     
>         p 'Creating api form'
>         api_namespace = ApiNamespace.create(params)
>         api_form_params =  { 
>             properties: { 
>                 "email": {
>                     "label": cta.email_label,
>                     "pattern": "",
>                     "required": "1",
>                     "field_type": "input",
>                     "prepopulate": "0",
>                     "type_validation": "email"
>                  }
>             },
>             api_namespace_id: api_namespace.id,
>             success_message: cta.success_message,
>             failure_message: cta.failure_message,
>             submit_button_label: cta.submit_button_label,
>             title: cta.title,
>             show_recaptcha: true,
>         }
>     
>         ApiForm.create(api_form_params)
>     
>         p 'creating api resources'
>     
>         cta.call_to_action_responses.each do |cta_response|
>            cta_response_params = {
>                     api_namespace_id: api_namespace.id,
>                     properties: cta_response.properties,
>                     created_at: cta_response.created_at,
>                     updated_at: cta_response.updated_at
>                 }
>     
>            api_resource =  ApiResource.create(cta_response_params)
>         end
>       end
> ```

This migration worked well in development! I made a small change to the field type validation. That said, I think we should create a mapping for the types in `app/models/api_form.rb`. Having dasherized/space separated strings is inconsistent. 

I suggest changing it to something like: 

``` ruby
INPUT_TYPE_MAPPING = {
  free_text: 'free text',
  number: 'number'
}
```

and using the symbols inside the application instead of hard coded strings.

donrestarone:
@Pralish here are the merge blockers: 

1. https://github.com/restarone/violet_rails/issues/256
2. https://github.com/restarone/violet_rails/issues/255

We are so close to getting this feature integrated!

app/models/api_action.rb
```diff
@@ -0,0 +1,78 @@
...
+      self.update(lifecycle_stage: 'complete', lifecycle_message: email)
+    rescue => e
+      self.update(lifecycle_stage: 'failed', lifecycle_message: e.message)
+      execute_error_actions
+    end
+  end
+
+  def send_web_request
+    begin
+      response = HTTParty.post(request_url,
```
donrestarone, line 41: @Pralish This should be settable, see https://github.com/restarone/violet_rails/issues/258

donrestarone:
@Pralish can we add some unit tests for these 2 blind spots? These tests will need to verify validations: email, presence etc

Screen Shot 2021-11-21 at 8 04 29 PM

Screen Shot 2021-11-21 at 8 05 25 PM

donrestarone:
@Pralish  once [this](https://github.com/restarone/violet_rails/pull/236#issuecomment-972399635)  and [this](https://github.com/restarone/violet_rails/pull/236#issuecomment-974983047) is solved we can merge!

Pralish:
>These tests will need to verify validations: email, presence etc

@donrestarone , Did you mean validations for dynamic property attributes?

donrestarone:
@Pralish yes, the validations for required fields, patterns, Email, numbers only etc. 

This also includes testing recapcha

db/migrate/20211118004240_migrate_call_to_actions_to_api.rb
```diff
@@ -0,0 +1,134 @@
...
+      api_namespace = ApiNamespace.create(params)
+      api_form_params =  { 
+          properties: { 
+              "name": {
+                  "label": cta.name_label,
+                  "pattern": "",
+                  "required": "0",
+                  "field_type": "input",
+                  "prepopulate": "0",
+                  "type_validation": "free text"
```
donrestarone, line 32: @Pralish this migration script should be updated. Since its only been run on staging we should be fine to update this one in-place.

lib/tasks/api_action.rake
```diff
@@ -0,0 +1,11 @@
...
+      ApiAction.where(lifecycle_stage: 'failed', action_type: ['send_email', 'send_web_request']).each do |action|
```
donrestarone, line 6: @Pralish you need to specify `tenant` here. Otherwise this will only run on the public schema. Rake tasks are global to the system.

donrestarone: @Pralish  see more about accessing all tenants in Rake tasks here: https://github.com/restarone/violet_rails/blob/master/lib/tasks/report.rake

lib/tasks/maintenance.rake
```diff
@@ -22,4 +22,10 @@ namespace :maintenance do
       end
     end
   end
+
+  task :clear_discarded_api_actions => [:environment] do
```
donrestarone, line 26: after fixing the tenant access issue: 
see more about accessing all tenants in Rake tasks here: https://github.com/restarone/violet_rails/blob/master/lib/tasks/report.rake

@Pralish this needs to be added to the `schedule.rb` run it once a day

PR #253, opened 2021-11-18
Data pipeline upgrade path: Remove CTA [MERGE AFTER FULL RELEASE]

Pralish:
This PR should only be merged after running the OTS and manually changing "{{cms:helper render_call_to_action :id}}" cta helper with api_form helper.


Pralish:
@donrestarone , This PR should only be merged after running the OTS and manually changing "{{cms:helper render_call_to_action :id}}" cta helper with api_form helper. 

Let me know what you think

donrestarone:
@Pralish I included the OTS as a migration here: https://github.com/restarone/violet_rails/pull/236/files#diff-22bdc0282cc9f8117de455983ef0639cfc22da575ba7a6cf5d1d077f45dcec1f

what are your thoughts?

app/services/analytics_report_service.rb
```diff
@@ -16,26 +16,12 @@ def analytics_report_json
...
       start_date: report_since.to_date,
       end_date: Date.today,
-      ctas: cta_json,
       visits: visits_json,
       users: users_json,
       macros: macros_json
     }
   end
 
-  def cta_json
```
donrestarone, line 26: @Pralish can you spec out/estimate how we can add analytics to API namespaces/resources? This is not a top priority but I'd like to have it done before the Violet Rails 1.0 release

donrestarone:
@Pralish I have notified clients that this deprecation is coming and they have all migrated onto API. Can you fix the merge conflicts and go over the migration strategy and a quick regression run to make sure we didn't miss anything? 

I'll deploy this to staging early next week.

Pralish:
> @Pralish I have notified clients that this deprecation is coming and they have all migrated onto API. Can you fix the merge conflicts and go over the migration strategy and a quick regression run to make sure we didn't miss anything?
> 
> I'll deploy this to staging early next week.

@donrestarone , Have they manually ran the migration script? Because the migration script didn't make any changes for me because of this condition when I ran the migration. 


> if models.include?("CallToAction")

Since we are removing the CTA model and running the migration in the same PR, the migration script will never execute because the CTA model won't be there anymore. Was this intentional?

Thanks

PR #266, opened 2021-11-24
Data pipeline feedbacks final

Pralish:


app/views/comfy/admin/api_forms/_form.html.haml
```diff
@@ -56,8 +56,8 @@
                     = a.select :field_type, options_for_select(['input', 'textarea'], properties[key]["field_type"] || 'input'), {}, {class: 'form-control' }
                   .form-group.col-12
                     = a.label :type_validation, 'Type validation', class: 'mr-4'
-                    = a.select :type_validation, options_for_select(ApiForm::INPUT_TYPE, properties[key]["type_validation"] || 'text'), {}, {class: 'form-control', onchange: "toggleRegex(this , '#{key}_regex')" }
-                  .form-group.col-12{style: "#{'display: none' unless properties[key]["type_validation"] == 'REGEX pattern' }", id: "#{key}_regex"}
+                    = a.select :type_validation, options_for_select(ApiForm::INPUT_TYPE_MAPPING.keys, properties[key]["type_validation"] || 'text'), {}, {class: 'form-control', onchange: "toggleRegex(this , '#{key}_regex')" }
+                  .form-group.col-12{style: "#{'display: none' unless properties[key]["type_validation"] == 'regex_pattern' }", id: "#{key}_regex"}
```
donrestarone, line 60: you should use the constant you defined [here](https://github.com/restarone/violet_rails/pull/266/files#diff-115eaea86e7fb2e74e1045aa541f956134ae663024878679d1a3e82731f53fd7) instead of hard coding `regex_pattern`

test/controllers/api/resource_controller_test.rb
```diff
@@ -123,6 +123,22 @@ class Api::ResourceControllerTest < ActionDispatch::IntegrationTest
...
+    payload = {
+      data: {
+        name: ''
+      }
+    }
+    assert_no_difference "@users_namespace.api_resources.count" do
+      post api_create_resource_url(version: @users_namespace.version, api_namespace: @users_namespace.slug), params: payload, headers: { 'Authorization': "Bearer #{api_client.bearer_token}" }
+    end
+    
+    assert_equal response.parsed_body["code"], 400
```
donrestarone, line 139: are we rendering the validation errors here? For example:

```json
HTTP/1.1 422 Unprocessable Entity
Content-Type: application/vnd.api+json

{
  "errors": [
    {
      "status": "422",
      "source": { "pointer": "/data/attributes/firstName" },
      "title":  "Invalid Attribute",
      "detail": "First name must contain at least three characters."
    }
  ]
}
```

test/controllers/resource_controller_test.rb
```diff
@@ -60,4 +60,55 @@ class ResourceControllerTest < ActionDispatch::IntegrationTest
...
+    payload = {
+      data: {
+          properties: {
+            first_name: 'Don',
+            last_name: 'Restarone'
+          }
+      }
+    }
+    assert_difference "@api_namespace.api_resources.count", +1 do
+      post api_namespace_resource_index_url(api_namespace_id: @api_namespace.id), params: payload
```
donrestarone, line 75: how do we stub the passing recapcha? It looks like the recapcha validation is only enforced on the client-side

donrestarone: I see the reason now, recapcha passes by default in test env: https://stackoverflow.com/questions/24011793/recaptcha-in-rails-verify-recaptcha-returning-always-true-in-test?rq=1

Pralish: Recaptcha is valid in test env by default. To test the invalid condition I had to add this:
```
Recaptcha.configuration.skip_verify_env.delete("test")
```

donrestarone: @Pralish yup! Makes sense now, thanks for the clarification

test/controllers/api/resource_controller_test.rb
Pralish: Yes we are, Updated the test case.Thanks

app/views/comfy/admin/api_forms/_form.html.haml
Pralish: Made the change, Thanks

donrestarone:
@Pralish is this ready to be merged?

Pralish:
@donrestarone, Yes its ready

PR #274, opened 2021-12-16
Data pipeline radio buttons

donrestarone:


donrestarone:
@Pralish This is the new base branch for the feature, we will merge after you add the end2end test for form submission with the radio buttons

Pralish:
@donrestarone , I am a bit confused. When you say end2end test, did you mean the controller tests that we already have or tests that simulate user interactions. Maybe with capybara or cypress

https://semaphoreci.com/community/tutorials/integration-testing-ruby-on-rails-with-minitest-and-capybara

donrestarone:
@Pralish I mistyped. We don't need to introduce a browser driven test framework. You can add your integration test with the same style here: https://github.com/restarone/violet_rails/blob/master/test/controllers/resource_controller_test.rb

We just need to make sure that we exercise the form submission with a api namespace that has radio buttons

PR #281, opened 2022-01-09
Deprecate cta

donrestarone:


Pralish:
@donrestarone , Did clients manually ran the migration script? The migration script didn't make any changes for me because of this condition when I ran the migration.

> if models.include?("CallToAction")

Since we are removing the CTA model and running the migration in the same PR, the migration script will never execute because the CTA model won't be there anymore. Was this intentional?

Thanks

donrestarone:
@Pralish I dont see the migration script on this PR, what happened to it?

Pralish:
@donrestarone, Looks like the migration script is already in master.

PR #287, opened 2022-01-16
Super user invitation bug

Pralish:


app/controllers/admin/subdomains_controller.rb
```diff
@@ -47,6 +47,19 @@ def invite_current_user_to_subdomain(subdomain)
     end
   end
 
+  def invite_current_super_user_to_subdomain(subdomain)
```
donrestarone, line 50: @Pralish you don't need to make a new method here, `invite_current_user_to_subdomain` already exists and you just need to modify that method to assign the new permissions

app/controllers/admin/subdomains_controller.rb
```diff
@@ -47,6 +47,19 @@ def invite_current_user_to_subdomain(subdomain)
...
+    Apartment::Tenant.switch subdomain.name do
+      User.invite!(
+        email: current_user.email, 
+        can_manage_web: true,
```
donrestarone, line 54: these permissions should be in a constant defined in the user model. It should be `CANONICAL_USER_PERMISSIONS` -- that way we just have to edit that constant when we add new permissions

test/controllers/admin/subdomains_controller_test.rb
```diff
@@ -120,6 +120,34 @@ class Admin::SubdomainsControllerTest < ActionDispatch::IntegrationTest
...
+        subdomain: {
+          name: name
+        }
+      }
+      perform_enqueued_jobs do
+        assert_changes "Devise::Mailer.deliveries.size" do
+          assert_difference "Subdomain.count", +1 do
+            post admin_subdomains_url, params: payload
+            Apartment::Tenant.switch name do
+              assert User.first
```
donrestarone, line 137: make sure to query the database only once by referencing user as `user = User.first`

Issue #288, opened 2022-01-16
page preview broken for apex domains


donrestarone:
When a user clicks on the link for a CMS page, the URL is set incorrectly. Subdomains work as expected. 
![Screenshot from 2022-01-16 11-47-59](https://user-images.githubusercontent.com/35935196/149669426-a30d1beb-9312-4afc-a58e-638b746ce485.png)

donrestarone:
@Pralish can you look into this?

donrestarone:
@Pralish any updates?

Pralish:
Hi Shashike,

Looking into it. Looks like the issue is coming from comfortable mexican
sofa. I will send a PR by Monday.

Thanks

PR #293, opened 2022-02-05
Page preview broken for apex domain

donrestarone:
https://github.com/restarone/violet_rails/issues/288


donrestarone:
@Pralish just tested on staging and the issue is still there. I tested under `www.restarone.solutions` as well as `restarone.solutions` the root page path was still resolved incorrectly to `https://www/`

Pralish:
@donrestarone , Did you run this script?

Apartment::Tenant.switch('public') do
  site = Comfy::Cms::Site.find_by(hostname: 'www')
  site.update(hostname: "[www.#{ENV['APP_HOST']}](http://www./#%7BENV%5B'APP_HOST'%5D%7D)") if site
end

If so can you give me the response of this:

Apartment::Tenant.switch('public') { Comfy::Cms::Site.all }

donrestarone:
@Pralish OTS is good, but we should have it as a migration

app/models/subdomain.rb
```diff
@@ -141,7 +141,7 @@ def destroy
 
   def self.unsafe_bootstrap_www_subdomain
     Apartment::Tenant.switch('public') do
-      Subdomain.bootstrap_via_comfy('public', 'www')
+      Subdomain.bootstrap_via_comfy('public', "www.#{ENV['APP_HOST']}")
```
donrestarone, line 144: @Pralish have you tried without the `www` ?

donrestarone: @Pralish any updates on this?

Pralish:
@donrestarone , I have made the changes. Please review and let me know if you have any questions.

db/migrate/20220317025428_update_bootstraped_subdomain_hostname.rb
```diff
@@ -0,0 +1,8 @@
+class UpdateBootstrapedSubdomainHostname < ActiveRecord::Migration[6.1]
+  def change
+    Apartment::Tenant.switch('public') do
```
donrestarone, line 3: @Pralish this should be in `up` and `down` blocks in case people have interesting configurations that might break if `www` is removed ;)

PR #297, opened 2022-02-10
Automate deployments with capistrano deploy

Pralish:


donrestarone:
@Pralish great progress, to wrap up we need to do the following

1. we need a way to get the IP address of the action and then allowlist it on the AWS security group
2. IP address should be passed dynamically (per project)

PR #301, opened 2022-02-21
RC: blue green deployments

donrestarone:


config/deploy/staging.rb
```diff
@@ -7,7 +7,7 @@
 # server "example.com", user: "deploy", roles: %w{app web}, other_property: :other_value
 # server "db.example.com", user: "deploy", roles: %w{db}
 
-server "restarone-solutions", user: 'ubuntu', roles: %w{app db web}
+server "54.163.245.12", user: 'ubuntu', roles: %w{app db web}
```
donrestarone, line 10: we should pass this value dynamically + allow the production deployment to be passed an IP address as well

.github/workflows/deploy.yml
```diff
@@ -0,0 +1,71 @@
...
+          bundler-cache: true 
+      - uses: actions/setup-node@v2
+        with:
+          node-version: 14.x
+      - name: Add public IP to AWS security group
+        uses: sohelamin/aws-security-group-add-ip-action@master
+        with:
+          aws-access-key-id: ${{ secrets.AWS_ACCESS_KEY_ID }}
+          aws-secret-access-key: ${{ secrets.AWS_SECRET_ACCESS_KEY }}
+          aws-region: 'us-east-1'
```
donrestarone, line 56: aws-region should be a variable and set dynamically

donrestarone:
@Pralish great progress on automated deployments! I have left some feedback on this PR. The next step would be to deploy to our production environments (ie; a client projects).

Here's the way I think it should work: 

1. add deploy key (ex; nikean-deploy)
2. Add environment (eg; `nikean.org`)
3. restrict deployment branch to main/master
4. Add AWS_ACCESS_KEY_ID, AWS_SECRET_ACCESS_KEY, AWS_SECURITY_GROUP_ID, AWS_REGION (to be added), DEPLOY_ENC_KEY, DEPLOY_ID_RSA_ENC, KNOWN_HOSTS and SSH_PRIVATE_KEY

let me know if that makes sense

Pralish:
@donrestarone , 
> we should pass this value dynamically
I didn't quite understand. When you say dynamically, do you mean read the IP from environment variable/github secrets?

Pralish:
> @Pralish great progress on automated deployments! I have left some feedback on this PR. The next step would be to deploy to our production environments (ie; a client projects).
> 
> Here's the way I think it should work:
> 
> 1. add deploy key (ex; nikean-deploy)
> 2. Add environment (eg; `nikean.org`)
> 3. restrict deployment branch to main/master
> 4. Add AWS_ACCESS_KEY_ID, AWS_SECRET_ACCESS_KEY, AWS_SECURITY_GROUP_ID, AWS_REGION (to be added), DEPLOY_ENC_KEY, DEPLOY_ID_RSA_ENC, KNOWN_HOSTS and SSH_PRIVATE_KEY
> 
> let me know if that makes sense

@donrestarone , To be clear, the requirement here is to deploy to all client servers at once, when we merge to main/master branch?

if that's the case, we might not have to create  environments for each client servers. We can pass list of servers in config/deploy/production.rb like this:

> server "nikean.org", user: 'ubuntu', roles: %w{app db web}
> server "sanjaysinghal.com", user: 'ubuntu', roles: %w{app db web}

and it will deploy to all those servers when we run `cap production deploy`

We can also pass host variable if we want to deploy to a single server:

> cap HOSTS=nikean.org production deploy

But, this might also not work in github action because of having to configure ssh and all. 

Let me know if that's the requirement and I will start working on this.

donrestarone:
@Pralish we have clients on different AWS accounts. We will need to set configurations on a per system basis. 

Eg: Nikean.org needs different SSH keys than violet.Restarone.solutions

PR #320, opened 2022-03-15
refactor editor logic to differentiate between ready only/edit mode

donrestarone:


donrestarone:
@Pralish can you get this over the finish line? We are trying to show a json editor in read only mode for read only views and in tree (edit) mode in edit views

Pralish:
@donrestarone , It looks like you have already worked on the feature. Let me know what else is to be done here.

PR #346, opened 2022-04-06
[Experimental] graphql support

donrestarone:


donrestarone:
did a basic implementation, so you can get lists like this: 
``` txt
{
  apiNamespaces {
    id
    name
    slug
    version
    properties
    apiResources {
      id
      properties
    }
  }
}

```
Screenshot from 2022-04-06 19-15-15

PR #353, opened 2022-04-13
[Experimental] External API connector 

donrestarone:
Adds external API connector that enables:

1. [Done] external services to be called with dynamic authentication, state tracking and error handling
2. [Done] run batch job with exponential backoff  (on the job itself)
3. [To do] run a job on demand or on schedule (cron)
4. [To do] implement editor for writing ruby code


donrestarone:
test is failing in CI but passing locally: 
``` bash
docker-compose run --rm solutions_test rails test test/models/external_api_client_test.rb
```

PR #432, opened 2022-04-20
WIP [API Plugin] Subdomain Events (#431)

donrestarone:
* add basic fixture for plugin_subdomain_events-- PS, you should never write tests that check for object.last unless you know for sure the last object is predictable -- ie; if you created it during the test run

* basic implementation + test to verify that it doesnt work when it shouldnt

* fix tests + add to UI

* hook into emailbox + track emails

* execute create api actions

* api action payload mapping issue

* fix nesting

* remove byebug

* works in development but this test is failing, need to explore further -- skipping for now

* refactor under test

* fix test

* add trackability to forum + add more tests


donrestarone:
need to fix: 

1. Email events are tracked, but actions are not executed (they get stuck in the initialized phase) 
2. No action runs correctly the first time (the first time forum post/thread runs the action fails, and the subsequent re-try will pass)

PR #433, opened 2022-04-20
[feature WIP] Add api plugin subdomain events

donrestarone:
need to fix: 

1. Email events are tracked, but actions are not executed (they get stuck in the initialized phase) 
2. No action runs correctly the first time (the first time forum post/thread runs the action fails, and the subsequent re-try will pass)


test/models/api_namespace_test.rb
```diff
@@ -1,7 +1,37 @@
...
+      assert_changes "@subdomain_events_api.create_api_actions.first.lifecycle_stage" do        
+        service.track_event
+        Sidekiq::Worker.drain_all
+      end
+    end
+    @subdomain_events_api.api_resources.reload.last.properties
+  end
+
+  test "plugin: subdomain/subdomain_events -> tracks message creation by creating ApiResource & running actions" do
+    skip('investigate why api_resource_id: is not set on the @subdomain_events_api.create_api_actions.first')
```
donrestarone, line 26: @Pralish any idea what's going on here?

Pralish: > @Pralish any idea what's going on here?

@donrestarone , Just checked the issue. 

It should be @subdomain_events_api.executed_api_actions.first 
OR
 if you want to be more specific
@subdomain_events_api.executed_api_actions.where(type: 'CreateApiAction').first
OR
@subdomain_events_api.api_resources.last.create_api_actions.first

@subdomain_events_api.create_api_actions.first is the template that api_resources will use to create the create_api_actions

Let me know if that makes sense

Issue #151, opened 2021-05-28
when file is attached in forum, notification email is broken 


donrestarone:
SimpleDiscussion::UserMailer#new_post 

ActionView::Template::Error: No route matches {:action=>"show", :controller=>"active_storage/blobs/redirect", :filename=>#<ActiveStorage::Filename:0x000055c2744b28a0 @filename="Violet Rails - Brand Gui...

``` txt
I, [2022-04-20T02:34:21.136350 #7618]  INFO -- : [ActiveJob] [ActionMailer::MailDeliveryJob] [dffce8b7-32af-4567-9749-12c8e422b390] Performing ActionMailer::MailDeliveryJob (Job ID: dffce8b7-32af-4567-9749-12c8e422b390) from Sidekiq(default) enqueued at 2022-04-18T23:31:08Z with arguments: "SimpleDiscussion::UserMailer", "new_post", "deliver_now", {:args=>[#<GlobalID:0x0000561f9fefb950 @uri=#<URI::GID gid://r-solutions/ForumPost/30>>, #<GlobalID:0x0000561f9fefa5f0 @uri=#<URI::GID gid://r-solutions/User/2>>]}
D, [2022-04-20T02:34:21.205045 #7618] DEBUG -- : [ActiveJob] [ActionMailer::MailDeliveryJob] [dffce8b7-32af-4567-9749-12c8e422b390]   ForumThread Load (1.7ms)  SELECT "forum_threads".* FROM "forum_threads" WHERE "forum_threads"."id" = $1 LIMIT $2  [["id", 3], ["LIMIT", 1]]
D, [2022-04-20T02:34:21.264447 #7618] DEBUG -- : [ActiveJob] [ActionMailer::MailDeliveryJob] [dffce8b7-32af-4567-9749-12c8e422b390]   Rendering simple_discussion/user_mailer/new_post.html.haml
D, [2022-04-20T02:34:21.276334 #7618] DEBUG -- : [ActiveJob] [ActionMailer::MailDeliveryJob] [dffce8b7-32af-4567-9749-12c8e422b390]   User Load (1.0ms)  SELECT "users".* FROM "users" WHERE "users"."id" = $1 LIMIT $2  [["id", 1], ["LIMIT", 1]]
[...]

PR #452, opened 2022-04-22
[bug fix] when file is attached in forum, notification email is broken

alis-khadka:
https://github.com/restarone/violet_rails/issues/151


app/views/active_storage/blobs/_blob.html.erb
```diff
@@ -10,7 +10,7 @@
       <span class="attachment__name"><%= blob.filename %></span>
       <span class="attachment__size"><%= number_to_human_size blob.byte_size %></span>
     <% end %>
-    <%= link_to rails_blob_url(blob, subdomain: Apartment::Tenant.current),  target: '_blank' do %>
+    <%= link_to Rails.application.routes.url_helpers.rails_blob_url(blob, subdomain: Apartment::Tenant.current, host: Rails.application.config.action_mailer.default_url_options[:host]),  target: '_blank' do %>
```
donrestarone, line 13: @alis-khadka if you're looking for the `host` take a look at using `ENV[APP_HOST]` here instead. Its better than using the action mailer default which is only initialized at boot time (ie; probably wont change & you want that)

test/mailers/simple_discussion/user_mailer_test.rb
```diff
@@ -0,0 +1,32 @@
+require "test_helper"
+
+class SimpleDiscussion::UserMailerTest < ActionMailer::TestCase
```
donrestarone, line 3: nice tests!

app/views/active_storage/blobs/_blob.html.erb
donrestarone: edit: I was mistaken. Initialization doesnt matter in this context (both the ENV and Rails.application.blah.blah are both boot time initialized)
But its a bit clearer when we use ENV[APP_HOST]:

``` ruby
ubuntu@ip-172-31-68-13:/var/www/violet/current$ rails c
RAILS_ENV=staging environment is not defined in config/webpacker.yml, falling back to production environment
Loading staging environment (Rails 6.1.5)
irb(main):001:0> ENV['APP_HOST']
=> "restarone.solutions"
irb(main):002:0> Rails.application.config.action_mailer.default_url_options[:host]
=> "restarone.solutions"
irb(main):003:0>
```

alis-khadka: Sure, I will push the update.

PR #459, opened 2022-04-23
Docker Dev workflow is broken (apple M1)

Pralish:
https://github.com/restarone/violet_rails/issues/402


donrestarone:
@Pralish any documentation on why this is the fix? Havnt tested yet obv

Pralish:
@donrestarone , Found the fix on stackoverflow.

https://stackoverflow.com/questions/70963924/unable-to-load-nokogiri-in-docker-container-on-m1-mac

Issue #468, opened 2022-04-23
rails image rendering broken in multiple systems


donrestarone:
<img width="692" alt="Screen Shot 2022-04-23 at 6 28 19 PM" src="https://user-images.githubusercontent.com/35935196/164948063-4bad25bc-dfd6-4061-8235-51643c65d4ef.png">

PR #471, opened 2022-04-24
[bug fix] minor change for forum email notification not sent when image file is attached

alis-khadka:
https://github.com/restarone/violet_rails/issues/468

### DIFF
--- a/app/views/active_storage/blobs/_blob.html.erb
+++ b/app/views/active_storage/blobs/_blob.html.erb
@@ -1,6 +1,6 @@
 <figure class="attachment attachment--<%= blob.representable? ? "preview" : "file" %> attachment--<%= blob.filename.extension %>">
   <% if blob.representable? %>
-    <%= image_tag Rails.application.routes.url_helpers.rails_representation_url(blob.representation(resize_to_limit: local_assigns[:in_gallery] ? [ 800, 600 ] : [ 1024, 768 ]).processed, host: ENV['APP_HOST']) %>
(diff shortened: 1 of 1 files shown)


app/views/active_storage/blobs/_blob.html.erb
```diff
@@ -1,6 +1,6 @@
 <figure class="attachment attachment--<%= blob.representable? ? "preview" : "file" %> attachment--<%= blob.filename.extension %>">
   <% if blob.representable? %>
-    <%= image_tag Rails.application.routes.url_helpers.rails_representation_url(blob.representation(resize_to_limit: local_assigns[:in_gallery] ? [ 800, 600 ] : [ 1024, 768 ]).processed, host: ENV['APP_HOST']) %>
+    <%= image_tag Rails.application.routes.url_helpers.rails_representation_url(blob.representation(resize_to_limit: local_assigns[:in_gallery] ? [ 800, 600 ] : [ 1024, 768 ]), host: ENV['APP_HOST']) %>
```
alis-khadka, line 3: @donrestarone 
The `processed` method creates and uploads the variant of attached image to the service before returning its url which is suggested not to be inline in a template. Removing `processed` will delay the processing (uploading to service).
Thus we can remove this.

reference: https://edgeapi.rubyonrails.org/classes/ActiveStorage/Variant.html
image

donrestarone: @alis-khadka sounds good, thanks for the quick fix.

PR #472, opened 2022-04-24
[bug fix] Images not rendering

donrestarone:
https://github.com/restarone/violet_rails/issues/468


donrestarone:
@alis-khadka This branch is on staging but it looks like the issue with the broken images is still there
Screenshot from 2022-04-23 22-25-42

alis-khadka:
I will look into this and update you

PR #501, opened 2022-05-03
[Violet Rails Core]: Event tracking

alis-khadka:


alis-khadka:
The `current_visit` variable was returning `nil` when the my current-session was long.
The ankane  team also suggests that `current_visit` might return `nil`. [Reference](https://github.com/ankane/ahoy/issues/320)

I have updated the `if clause condition` to `if Subdomain.current.tracking_enabled && current_visit`.

test/controllers/admin/comfy/pages_controller_test.rb
```diff
@@ -2,8 +2,8 @@
 
 class Comfy::Admin::Cms::PagesControllerTest < ActionDispatch::IntegrationTest
   setup do
-    @restarone_subdomain = Subdomain.find_by(name: 'restarone').name
-    Apartment::Tenant.switch @restarone_subdomain do
+    @restarone_subdomain = Subdomain.find_by(name: 'restarone')
```
donrestarone, line 5: thanks for the test clean up! @alis-khadka

PR #359, opened 2022-04-14
[Experimental] Event tracking

donrestarone:
to do: 

1. track when user views forum post
2. track when user edits forum post
3. track when user edits a CMS page
4. track when a user edits a file
5. track when a user edits a user
6. track when a user views an email

Show events under users
Show events under ahoy visits

PR #511, opened 2022-05-03
fixed the width of flash message

Gloria-kuang:
The flash message is now showing in full width under the navbar.
Screen Shot 2022-05-03 at 8 11 26 PM


donrestarone:
@Gloria-kuang welcome & thanks for the bug fix!

Your change is due to go out after we release https://github.com/restarone/violet_rails/pull/359 (which is in turn currently [blocked by our production deployment](https://github.com/restarone/violet_rails/actions/runs/2266696340))

While we wait, can you update your description with a picture or some other context for posterity?

PR #517, opened 2022-05-03
[Bug fix] enforce ruby hashes in JSONB columns

donrestarone:


donrestarone:
@Pralish thanks for the quick turnaround on the RCA and fix. Looks like the new migration ran again on staging-- and the problem is in the Violet subdomain: 
``` bash
ubuntu@ip-172-31-86-57:/var/www/violet/current$ cat log/parse_json_string_error.log 
# Logfile created on 2022-05-03 22:15:59 +0000 by logger.rb/66358
E, [2022-05-03T22:15:59.801109 #460287] ERROR -- : Tenant: violet, Entity: ApiNamespace, Object Id: 6, Field: properties, Field Value: model_definition {"record_id"=>{"type"=>"integer", "validations"=>{"unique"=>true, "allow_blank"=>false, "primary_key"=>true}}, "record_type"=>{"type"=>"string", "validations"=>{"allow_blank"=>false}}}
ubuntu@ip-172-31-86-57:/var/www/violet/current$

```

Issue #57, opened 2021-04-16
comfymexicansofa can't create categories 


donrestarone:
creating categories/ajax actions result in a blank page but the category is created (the Ajax action is broken)

Pralish:
## RCA

#### Issue:
creating categories/ajax actions result in a blank page but the category is created (the Ajax action is broken)

#### Cause:
Action view was not creating remote form

##### Expected:
<img width="529" alt="Screen Shot 2022-05-04 at 10 21 32 AM" src="https://user-images.githubusercontent.com/50227291/166623755-7b969ef8-b8e0-4695-ae1f-f3f5a1045218.png">

##### Actual:
<img width="529" alt="Screen Shot 2022-05-04 at 10 22 05 AM" src="https://user-images.githubusercontent.com/50227291/166623774-f52004b4-4923-4139-b284-c87cf19eb520.png">

Note that the 'data-remote' attribute is missing

##### Background:

comfortable_mexican_sofa uses 'form_with' helper to generate categories form. 'form_with' helper checks this config to generate remote form.
> config.action_view.form_with_generates_remote_forms

This config was by default true, so a remote form was expected.

As of Rails 6, the default is false. 

- https://discuss.rubyonrails.org/t/rails-6-1-remote-forms-are-no-longer-default/76912

- https://github.com/rails/rails/commit/3aef59cae2a4b4398b80c78d49361fa64e05c0fd#diff-44b94eca66c7497711821a8e6bcdfde4684bb7b8efa15e64da6532449f03ef0b

#### Fix:
- Either add 'local: false' to the form
- OR update the config [Implemented]

https://github.com/restarone/violet_rails/pull/520/

PR #520, opened 2022-05-04
fix(): Fix categories form

Pralish:
https://github.com/restarone/violet_rails/issues/57

### DIFF
--- a/config/application.rb
+++ b/config/application.rb
@@ -29,5 +29,8 @@ module RSolutions
 
     # since web console usage depends on subdomain rules
     config.web_console.development_only = false
+
+    # generates remote forms with form_with 
+    config.action_view.form_with_generates_remote_forms = true
   end
 end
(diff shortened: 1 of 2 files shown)


config/application.rb
```diff
@@ -29,5 +29,8 @@ class Application < Rails::Application
 
     # since web console usage depends on subdomain rules
     config.web_console.development_only = false
+
+    # generates remote forms with form_with 
+    config.action_view.form_with_generates_remote_forms = true
```
donrestarone, line 34: @Pralish you mentioned that the fix was to:

    Either add 'local: false' to the form
    OR update the config [Implemented]

Can we only add the local flag? I feel like setting the config could block us from Rails upgrades down the road

Pralish: @donrestarone , We will have to add local flag to comfertable_mexican_sofa repo. Is that ok?

donrestarone: @Pralish yeah lets do that and add a test for the redirect!

donrestarone: no hurry, take your time on the change since we have a queue to drain @Pralish

Issue #455, opened 2022-04-22
when nav bar is fixed flash messages cannot be displayed in CMS pages/layout


donrestarone:
<img width="1722" alt="Screen Shot 2022-04-22 at 3 26 41 PM" src="https://user-images.githubusercontent.com/35935196/164781366-d1f6f912-4e4e-4c54-80d5-9035f3456541.png">

https://user-images.githubusercontent.com/35935196/165644363-59fdf139-98c7-41f8-9e0d-c1e10c7ec200.mov

PR #524, opened 2022-05-06
fix(): #455 flash message needs to be compatible with custom navbar/CSS (#522)

donrestarone:
addresses https://github.com/restarone/violet_rails/issues/455
* fix issue where flash message was not showing as expected for violet.restarone.solutions/challenge.restarone.solutions


donrestarone:
tested on `violet.restarone.solutions` and `challenge.restarone.solutions` we see that the flash message is not clearing the navbar in the case of `violet.restarone.solutions`-- @alis-khadka  

Screenshot from 2022-05-06 19-44-50
Screenshot from 2022-05-06 19-42-11

PR #550, opened 2022-05-12
add middleware for gzip (#549)

donrestarone:
add middleware for gzip (#549)


donrestarone:
tested on `restarone.solutions`, looks like this is unnecessary

Issue #559, opened 2022-05-13
[API Namespace - form rendering] Select element has an extra empty option


Ayon95:
The HTML `<select>` element gets rendered with an extra empty `<option>` element.

![extra-option-element](https://user-images.githubusercontent.com/73725297/168395283-4ca30300-d9d6-4363-b40a-648235ed8b69.png)
![select-options](https://user-images.githubusercontent.com/73725297/168395284-26aa9a94-58b9-475d-ba7d-4bd35ce7e09e.png)
![select-ui](https://user-images.githubusercontent.com/73725297/168395286-6634112b-bcee-456a-a3b6-7805b9130081.png)

donrestarone:
@Ayon95 can you attach an image of your form configuration too?

Ayon95:
@donrestarone you mean this?
![gender-select-input-config](https://user-images.githubusercontent.com/73725297/168396000-6c3f14d8-97e4-4aea-b2c9-fb0b56f060d6.PNG)

donrestarone:
@Ayon95 yup! That's it, thanks

donrestarone:
@Ayon95 @Pralish this is deprioritized, lets take care of this after this current release.

PR #562, opened 2022-05-14
 spike fix api form select has extra empty option

donrestarone:
spike for https://github.com/restarone/violet_rails/issues/559

@Pralish since you have the most context in this system can you look at this issue? We need to make sure that we can control if a blank option shows in the single/multi select. 

CC @Ayon95 in case you need to explain the issue over discord

### DIFF
--- a/app/helpers/api_forms_helper.rb
+++ b/app/helpers/api_forms_helper.rb
@@ -16,9 +16,40 @@ module ApiFormsHelper
         render partial: 'comfy/admin/api_forms/radio', locals: {form: form, key: key, value: value, form_properties: form_properties}
       else
         if form_properties[key]['select_type'] == 'multi' || !value.present?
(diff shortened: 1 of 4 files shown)


Pralish:
@donrestarone , We have a weird issue with select tags. The blank option is needed if we want to make the field non mandatory, clear current selection or for placeholder. 

The empty option comes from 'include_blank' attribute in this line
https://github.com/restarone/violet_rails/blob/6acea8e0ec46ab33a7752cefe1a7877d849d9876/app/helpers/api_forms_helper.rb#L21

We can remove the empty option, but that means the first option is selected by default. We cannot clear the selection [must select one of the option]

 @Ayon95 ,  Do we have any suggestion how to add placeholders/empty option in select tag?
 
 Ref: 
 https://stackoverflow.com/questions/5805059/how-do-i-make-a-placeholder-for-a-select-box
 https://stackoverflow.com/questions/4557701/ruby-on-rails-how-do-i-use-a-default-placeholder-thing-in-a-select-tag

donrestarone:
@Pralish 

I believe the code is working as expected, its only a missing feature (since we are trying to use radio buttons as select) dont resolve `include_blank: true` more intelligently. We should do something like, 

1. if required: it should look for a prepopulated value, if present -- dont do include blank / 
2. if not required, include a blank 

But I think we should put this task on pause since we have other more pressing tasks in [this upcoming release](https://github.com/restarone/violet_rails/issues?q=is%3Aissue+is%3Aopen+label%3A%22API+-+data+pipeline+v2%22)

Ayon95:
@Pralish Hey, I think I can use the first blank `option` as placeholder text for the `select` element.

Issue #555, opened 2022-05-13
[API Namespace] cannot change the order of properties of an object


Ayon95:
Properties of an object cannot be rearranged. They seem to get rearranged back to their initial order after hitting Save.

Suppose, I want the properties to be arranged in this order:
![Capture](https://user-images.githubusercontent.com/73725297/168186271-7d5c8002-9a9d-4c48-8533-0fec017cf0ec.PNG)

The new order doesn't get applied after saving:
![Capture2](https://user-images.githubusercontent.com/73725297/168186332-d7c4fce5-23c1-4913-bbe7-97b36484adde.PNG)

donrestarone:
@Pralish has the most context on this

donrestarone:
we see that the issue is not in the client / params/ controller because the object property order is preserved while it is in the controller layer: 
``` txt
[42, 51] in /var/app/app/controllers/comfy/admin/api_namespaces_controller.rb
   42:
   43:   # PATCH/PUT /api_namespaces/1 or /api_namespaces/1.json
   44:   def update
   45:     respond_to do |format|
   46:       byebug
=> 47:       if @api_namespace.update(api_namespace_params)
   48:         format.html { handle_success_redirect }
   49:         format.json { render :show, status: :ok, location: @api_namespace }
   50:       else
   51:         format.html { handle_error_redirect }
(byebug) params
[...]

PR #564, opened 2022-05-14
spike fix order of properties for API Form

donrestarone:
https://github.com/restarone/violet_rails/issues/555

@Pralish  @alis-khadka  I have added some helpful breakpoints so yall can debug the above issue: 

when you change the order of the attributes in the API namespace here and hit save -- the breakpoints should get hit: 
Screenshot from 2022-05-13 22-17-10


donrestarone:
@Pralish have you had some time to think about this one? I believe the issue is not in the params (the shape of the object is preserved when transmitted to the server)

Issue #527, opened 2022-05-07
ability to define a custom Api action  similar to external Api client


donrestarone:
see example here: https://github.com/restarone/violet_rails/wiki/API:-Entities,-Form-Rendering,-Interfaces-and-Actions#automation-connecting-and-extracting-data-from-external-systemsapis

donrestarone:
instead of defining a whole class, it should just be an override to a method (say, `run_custom_action`) that is passed the following parameters: 

current_user (if any), current visit,  Api action/namespace/resource etc

donrestarone:
example implementation: 
```ruby
class CustomAction
  def run_custom_action(current_user = nil)
    raise "not implemented error"
  end 
end
```

donrestarone:
@alis-khadka when you get a chance can you open a PR with just this feature + the bug fixes against master? Thanks :pray: 
due to time constraints we removed your feature from this week's release: https://github.com/restarone/violet_rails/pull/589

it will be shipping in the future instead. So its good to have a fresh PR targeting master in the next few days.

PR #569, opened 2022-05-14
[violet rails core] Ability to define a custom API action (ruby method) similar to External Api Client

alis-khadka:
https://github.com/restarone/violet_rails/issues/527

# Acceptance criteria
when deployed to `staging`, it should

1. allow N number of custom actions to be performed 
2. log errors in the response (similar to HTTP action)
3. handle custom actions in the controller context
4. handle custom actions out of the controller context (ie; sending an email)
5. not break existing API action workflows

### DIFF
--- a/app/controllers/concerns/api_actionable.rb
+++ b/app/controllers/concerns/api_actionable.rb
@@ -37,21 +37,18 @@ module ApiActionable
     flash[:notice] = @api_namespace.api_form.success_message
     if @custom_action.present?
       begin
-        custom_api_action == ApiAction::CustomApiAction.new
+        custom_api_action = CustomApiAction.new
 
         eval("def custom_api_action.run_custom_action(api_action: , api_namespace:, api_resource: , current_visit: , current_user: nil); #{@custom_action.method_definition};end;")
-
         response = custom_api_action.run_custom_action(api_action: @custom_action, api_namespace: @api_namespace, api_resource: @api_resource, current_visit: current_visit, current_user: current_user)
 
         # TO DO: what should we use for lifecycle_message
-        self.update(lifecycle_stage: 'complete', lifecycle_message: response.to_s)
+        @custom_action.update(lifecycle_stage: 'complete', lifecycle_message: response.to_s)
       rescue => e
-        self.update(lifecycle_stage: 'failed', lifecycle_message: e.message)
+        @custom_action.update(lifecycle_stage: 'failed', lifecycle_message: e.message)
         execute_error_actions
       end
     end
-
-    #redirect_back(fallback_location: root_path, notice: "Api resource was successfully updated.")
   end
 
   def handle_redirection
--- a/app/controllers/concerns/api_actionable.rb
+++ b/app/controllers/concerns/api_actionable.rb
@@ -2,12 +2,17 @@ module ApiActionable
   extend ActiveSupport::Concern
   included do
     before_action :initialize_api_actions, only: [:update, :show, :destroy]
+    before_action :check_for_custom_action, only: [:create, :update, :show, :destroy]
     before_action :check_for_redirect_action, only: [:create, :update, :show, :destroy]
     after_action :execute_api_actions, only: [:show, :create, :update, :destroy]
     before_action :check_for_serve_file_action, only: [:show, :create, :update, :destroy]
     rescue_from StandardError, with: :handle_error
   end
 
+
+  def check_for_custom_action
+    @custom_action = @api_namespace.send(api_action_name).where(action_type: 'custom_action').last
+  end
 
   def check_for_redirect_action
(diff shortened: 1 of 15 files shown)


app/controllers/concerns/api_actionable.rb
```diff
@@ -28,6 +33,24 @@ def check_for_serve_file_action
...
 
+  def handle_custom_action
+    flash[:notice] = @api_namespace.api_form.success_message
+    if @custom_action.present?
+      begin
+        custom_api_action = CustomApiAction.new
+
+        eval("def custom_api_action.run_custom_action(api_action: , api_namespace:, api_resource: , current_visit: , current_user: nil); #{@custom_action.method_definition};end;")
+        response = custom_api_action.run_custom_action(api_action: @custom_action, api_namespace: @api_namespace, api_resource: @api_resource, current_visit: current_visit, current_user: current_user)
+
```
donrestarone, line 44: you should stringify the response so its possible to return something like: 

``` ruby
{result: 'success', code: 200}
# or
"everything went well"
```

app/controllers/concerns/api_actionable.rb
```diff
@@ -28,6 +33,24 @@ def check_for_serve_file_action
     serve_file_action.update(lifecycle_stage: 'complete', lifecycle_message: "label: #{file.label} id: #{file.id} mime_type: #{file.attachment.content_type}")
   end
 
+  def handle_custom_action
+    flash[:notice] = @api_namespace.api_form.success_message
+    if @custom_action.present?
+      begin
+        custom_api_action = CustomApiAction.new
+
+        eval("def custom_api_action.run_custom_action(api_action: , api_namespace:, api_resource: , current_visit: , current_user: nil); #{@custom_action.method_definition};end;")
```
donrestarone, line 42: @alis-khadka @Pralish  we should create a ticket that scrubs method/model definitions coming into Violet Rails of things like: `Subdomain`, `Apartment`, `Tenant`, `switch`, `SomeModel.destroy_all` -- so we can open this upto platform developers on subdomains instead of whole systems (like how we have it now, because of this vulnerability)

app/helpers/application_helper.rb
```diff
@@ -16,7 +16,7 @@ def page_entries_info(collection)
   def execute_actions(resource, class_name)
     api_actions = resource.send(class_name)
     api_actions.each do |api_action|
-      api_action.execute_action unless api_action.serve_file? || api_action.redirect?
+      api_action.execute_action unless api_action.serve_file? || api_action.redirect? || api_action.custom_action?
```
donrestarone, line 19: we should document that in the order of operations, custom actions should come last! @alis-khadka

app/models/custom_api_action.rb
```diff
@@ -0,0 +1,5 @@
+class CustomApiAction
+  def run_custom_action(api_action: , api_namespace:, api_resource: , current_visit: , current_user: nil)
+    raise "not implemented error"
```
donrestarone, line 3: nice API @alis-khadka

test/controllers/resource_controller_test.rb
```diff
@@ -155,4 +155,26 @@ class ResourceControllerTest < ActionDispatch::IntegrationTest
...
+    api_action.update!(method_definition: "User.create!(email: 'contact1@restarone.com', password: '123456', password_confirmation: '123456', confirmed_at: Time.now)")
+
+    payload = {
+      data: {
+          properties: {
+            name: 123,
+          }
+      }
+    }
+    assert_difference "api_namespace.api_resources.count", +1 do
```
donrestarone, line 173: nice tests @alis-khadka

test/fixtures/api_actions.yml
```diff
@@ -100,4 +100,12 @@ create_api_action_plugin_subdomain_events:
   api_namespace: plugin_subdomain_events
   http_method: post
 
+create_custom_api_action_one:
+  type: CreateApiAction
+  action_type: custom_action
+  include_api_resource_data: false
+  payload_mapping:
+  method_definition: ""
```
donrestarone, line 108: @alis-khadka you should include a basic example of something here, like writing to the logs (API action response logs) or something like that

donrestarone:
@alis-khadka great first stab, a few things to note: 

1. This PR should target RC (we are currently prepping a release candidate for [Data pipelines V2](https://github.com/restarone/violet_rails/issues?q=is%3Aissue+is%3Aopen+label%3A%22API+-+data+pipeline+v2%22) )
2. please add a test that runs N number (at least 3 discrete routines) of custom API actions-- we need to mathematically validate that we can process those custom tasks in order (and save the response as expected)-- provided that the action author wrote them correctly

app/helpers/application_helper.rb
alis-khadka: I did not understand it clearly

PR #585, opened 2022-05-17
Change order of api actions execution

alis-khadka:


app/controllers/comfy/admin/api_resources_controller.rb
```diff
@@ -31,18 +29,12 @@ def create
...
 
-        format.html {
-          if @redirect_action.present?
-            handle_redirection
-          else
-            redirect_to api_namespace_resource_path(api_namespace_id: @api_resource.api_namespace_id,id: @api_resource.id), notice: "Api resource was successfully created."
-          end
-        }
+        # Preventing double render error
+        return if @redirect_action.present?
```
alis-khadka, line 35: Since, we are executing the api_actions in a defined order, we need to prevent the double render error.

app/controllers/concerns/api_actionable.rb
```diff
@@ -58,8 +57,8 @@ def handle_custom_actions
 
           eval("def custom_api_action.run_custom_action(api_action: , api_namespace:, api_resource: , current_visit: , current_user: nil); #{custom_action.method_definition};end;")
           response = custom_api_action.run_custom_action(api_action: custom_action, api_namespace: @api_namespace, api_resource: @api_resource, current_visit: current_visit, current_user: current_user)
-          response = response.present? ? response.as_json : "everything went well."
-          custom_action.update(lifecycle_stage: 'complete', lifecycle_message: response)
+
+          custom_action.update(lifecycle_stage: 'complete', lifecycle_message: response.to_json)
```
donrestarone, line 61: is there a reason for doing `response.to_json` ? 

What if the implicit return value was a `boolean` or `string` ? 

A better approach would just be to do `response.to_s` or just `response`-- any opinions on this @Pralish ?

Pralish: @donrestarone, converting boolean or string or even nil to json won't raise any errors. I prefer to_json because we can parse it easily if you ever wish to use the response.

But if we wish to just see the response, to_s looks cleaner.

donrestarone: @alis-khadka I stand corrected! 

Thanks for the clarification @Pralish -- this doesn't need to be changed in that case

PR #591, opened 2022-05-19
Custom html snippet for flash message

Pralish:


donrestarone:
@Pralish can we add the following to the spec? 

1. expose the newly created api_resource in the snippet context. So we can do the following: 

``` ruby
<h1>Thanks for filling out the form #{ api_resource.properties[:first_name] }. We will get back to you soon!</h1>
```

app/helpers/api_forms_helper.rb
```diff
@@ -70,5 +70,19 @@ def map_form_field_type(data_type)
...
+
+  # custom_message format: api_resource.properties["first_name"]
+  def parse_snippet(custom_snippet, api_resource)
+    return unless custom_snippet.present?
+
+    parsed_snippet = custom_snippet
+    # couldn't gsub directly because of escape characters added by ruby
+    # eval failed on "\#{api_resource.properties[\"String\"]}"
+    custom_snippet.scan(/\#\{(.*?)\}/).each do |code|
+      # TODO: sanitize code before eval
```
donrestarone, line 82: @Pralish After we merge https://github.com/restarone/violet_rails/pull/625 can we hook up the concern to this ?

PR #592, opened 2022-05-19
Event tracking rev 2

alis-khadka:
https://user-images.githubusercontent.com/35935196/169381648-481196d0-1077-4909-9d00-c89f1130a8f6.mp4

tracking-page-1


donrestarone:
@alis-khadka this is targeting the wrong branch. Currently its targeting RC -- it should target master (and you should remove the commits from your previous [feature which we tabled](https://github.com/restarone/violet_rails/issues/527))

PR #594, opened 2022-05-19
[bug fix] comfymexicansofa can't create categories

donrestarone:
https://github.com/restarone/violet_rails/issues/57

# acceptance criteria: 

when deployed to staging it should: 

1. allow for CMS categories to be created across all CMS entities


donrestarone:
after we test the patch on staging we will need to merge the change to master in the upstream repo (https://github.com/restarone/comfortable-mexican-sofa) and then create a new release tag (version 3.4 in this case)

CC @Pralish

donrestarone:
changes look good on staging, we should bump the version in the CMS repo

Issue #563, opened 2022-05-14
[API] #show view covers edit / back buttons on certain screen sizes


donrestarone:
we can see that on certain screen sizes the edit | back buttons on the UI are not visible (they are covered by the navbar)

![Screenshot from 2022-05-13 21-37-56](https://user-images.githubusercontent.com/35935196/168406120-6b50b805-7fe7-42db-ac7b-b5795b762ba4.png)
![Screenshot from 2022-05-13 21-37-54](https://user-images.githubusercontent.com/35935196/168406121-073f7e14-d1be-4579-ade5-34f276c07bb6.png)
![Screenshot from 2022-05-13 21-37-26](https://user-images.githubusercontent.com/35935196/168406123-09099bd4-5425-4c68-ac2d-85bc45b75246.png)

PR #611, opened 2022-05-27
[bug fix] "Add logo" shifting navbar down on mobile (restarone#563)

yashkir:
* add same classes to the link as to the image to hide it on mobile #563


donrestarone:
Thanks for your contribution @yashkir ! This is getting packaged into the next Violet Rails release for testing

Issue #600, opened 2022-05-20
[forum] when a forum thread/post with an image is updated, the thumbnail is broken + user is presented with error


donrestarone:
``` txt

An Errno::ENOENT occurred in forum_posts#update:

  No such file or directory @ rb_file_s_size - /tmp/image_processing20220509-1111367-6zqr2f.png
  app/controllers/simple_discussion/forum_posts_controller.rb:25:in `update'


```

see more info here: https://support.restarone.solutions/mailbox/message_threads/15

PR #619, opened 2022-05-30
[Bug fix] When a forum post is updated with with image/file, user is presented with error and thumbnail is broken for images.

donrestarone:
addresses: https://github.com/restarone/violet_rails/issues/600

# acceptance criteria

1. existing forum posts/threads that contain assets (images, files etc) are not broken
2. user is able to update posts/threads with/without assets 
3. functionality works as expected in subdomains and the apex domain


donrestarone:
@alis-khadka tested on `mushfiqur.restarone.solutions` and `restarone.solutions` and the patch works

we will let it 'simmer' for a little while so we can monitor notification email rendering and how automation workflows react to this change. 

I am slating this to go GA tomorrow at 5PM EST.

As always, Thanks for your contribution!

PR #625, opened 2022-06-01
Sanitize method/model definitions coming into Violet Rails

Pralish:


donrestarone:
@Pralish any updates on this?

PR #634, opened 2022-06-02
[Feature] Deploy to Heroku  (#633)

donrestarone:
addresses https://github.com/restarone/violet_rails/issues/385

Co-authored-by: Prashant <alish.khadka@gmail.com>
Co-authored-by: donrestarone <shashikejayatunge@gmail.co>

PR #635, opened 2022-06-02
Heroku deployment: fix node.js

donrestarone:


donrestarone:
@alis-khadka please use this as the base branch for addressing: https://github.com/restarone/violet_rails/pull/634

donrestarone:
@alis-khadka can we try pushing to heroku with the same node version we run in production?

alis-khadka:
> @alis-khadka can we try pushing to heroku with the same node version we run in production? 

Sure, I will be looking into this for today.

Issue #560, opened 2022-05-13
[API Namespace - flash message] feature to define custom HTML snippet for flash message content + fix close button positioning


Ayon95:
Often times, it is necessary to have custom formatting and styles for the content of flash message. So, it'd be nice if a user of Violet Rails could define their own HTML snippet from "Editing api_form" UI. The `<strong></strong>` element can be removed, and the HTML snippet can be inserted into `<div class="toast-body"></div>`.

Currently, only the flash message text can be set which gets injected into `<div class="toast-body"><strong></strong></div>`. So, there would be two options - either specify just a body of text, or custom HTML snippet.

Setting flash message text:
![setting-flash-message-text](https://user-images.githubusercontent.com/73725297/168398322-08f767d3-d738-4cc3-b223-155474665820.PNG)

Flash message markup:
![flash-message-markup](https://user-images.githubusercontent.com/73725297/168398383-01653055-a09f-47dd-a2f0-d7488c3c09f7.png)

Also, the `btn-close` button is centered vertically which looks odd when there is a lot of text. Its position can be fixed by adding the following styles to it:
```css
.btn-close {
   align-self: start;
   margin-top: 10px;
}
```
This is what the close button looks like after applying the above styles:
![flash-message-close-button-fixed](https://user-images.githubusercontent.com/73725297/168398247-5fa40abb-8a13-476f-8c1d-98ef97a810b5.PNG)

Without the above styles:
![flash-message-appearance](https://user-images.githubusercontent.com/73725297/168398281-4f0747c6-43fe-463d-b3df-beef37f82799.PNG)

PR #662, opened 2022-06-04
[Feature] Custom html snippet for flash message (#591)

donrestarone:
solves: https://github.com/restarone/violet_rails/issues/560

## [Feature] Custom html snippet for flash message

Dynamic segments are in flash messages! 🎉 Now you can,

1. use ERB / string interpolation in messages for dynamic segments 
2. Optionally use HTML, CSS, JS if you need heavily customized flash notification

## Screenshots
Bootstrap HTML with ERB example: 
Screen Shot 2022-07-26 at 7 17 56 PM

rendered: 
Screen Shot 2022-07-26 at 7 12 43 PM

## Demo

https://user-images.githubusercontent.com/35935196/180994051-e1753ea7-444d-4631-9b2d-b6d11ea6ae3f.mov


donrestarone:
@Pralish I'm deploying your branch to Heroku for alpha testing, let me know which order you want to merge your work with regards to this PR: https://github.com/restarone/violet_rails/pull/625

donrestarone:
@Pralish custom code works! But when I try to access the `api_resource` explicitly as `api_resource` or `self`-- its erroring: 
Screen Shot 2022-06-04 at 7 05 20 PM

donrestarone:
@Pralish any updates on this?

donrestarone:
@Pralish Thanks for fixing merge conflicts, we can start UAT once you get CI passing

app/models/api_resource.rb
```diff
@@ -2,6 +2,10 @@ class ApiResource < ApplicationRecord
   include JsonbFieldsParsable
 
   after_initialize :inherit_properties_from_parent
+  
+  after_initialize do
```
restarone-solutions, line 6: @Pralish 
Any specific reason for using `after_initialize` ? Because it gets called even for `Model.new`. We could use `after_find` which gets called only after a record has been fetched from the database.

alis-khadka, line 6: @Pralish 
Any specific reason for using `after_initialize`? Because it gets called even for `Model.new`. We could use `after_find` which is invoked only after the record has been fetched from database.

donrestarone:
Once @Pralish addresses @alis-khadka 's comment here: https://github.com/restarone/violet_rails/pull/662#discussion_r919855553 we will need to add in-line documentation as well

donrestarone:
## Heroku UAT  :test_tube: 
Error HTML Snippet should support
1.  interpolating `current_user`, `current_visit`( if any) + always `api_resource` 
2. plain HTML + CSS 
3. inline javascript (just like a snippet)

donrestarone:
@alis-khadka @Pralish please coordinate amongst the two of you to fix merge conflicts on this branch and [address the UAT findings here](https://github.com/restarone/violet_rails/pull/662#issuecomment-1185054052) (HTML and JS cant be used in HTML snippet)

app/models/api_resource.rb
Pralish: @alis-khadka , It's because we need that for new records as well.

Pralish:
@donrestarone, I have addressed the issues. Thanks

app/views/comfy/admin/api_actions/_form.html.haml
```diff
@@ -115,6 +115,9 @@
         = label_tag 'Subject'
         =  text_field_tag "api_namespace[#{type}_attributes][#{index}][email_subject]", resource.email_subject, { class: "form-control form-control-sm" }
 
+      .form-group
```
donrestarone, line 118: @Pralish is this change necessary? At the moment, I don't understand how its related to the feature

donrestarone:
@Pralish I think we need more test coverage on this feature to cover flows like: 

1. ERB in flash message  (should parse dynamic segments)
2. custom HTML in flash message (should show custom HTML)
3. string only in flash message (should show default toast)

PR #664, opened 2022-06-06
[feature] custom API Actions

donrestarone:
## Custom API Actions

Screen Shot 2022-07-18 at 6 14 50 PM

Screen Shot 2022-07-18 at 5 57 50 PM


donrestarone:
@alis-khadka if/when we work on this again -- please use this as the base branch

donrestarone:
@alis-khadka 

## 🧪  UAT findings

### 1. HTTP request return value is not the same as before

Screen Shot 2022-07-18 at 5 52 13 PM


Screen Shot 2022-07-18 at 5 53 29 PM
Screen Shot 2022-07-18 at 5 53 23 PM



#### 2. Custom actions should fail silently (to the user)
When an action fails its great that we are logging the failure.
Screen Shot 2022-07-18 at 6 07 51 PM


but it shouldn't result in a 500 to the user who submits the form: 

Screen Shot 2022-07-18 at 6 07 54 PM

donrestarone:
@alis-khadka was the issue with the HTTP response value patched? 
I still see the issue after your commits: 
Screen Shot 2022-07-19 at 9 06 10 PM

alis-khadka:
> @alis-khadka was the issue with the HTTP response value patched? I still see the issue after your commits: Screen Shot 2022-07-19 at 9 06 10 PM
> 
> This is the only blocker, besides that it LGTM

Previously, I misunderstood the UAT finding the other way. Apologies for this
I have pushed the changes for this. 
image

donrestarone:
@alis-khadka after that commit it is showing as `null`

Could you write a unit test that covers this use case? We should patch against persisted values changing (that is technically data corruption when the system starts saving different things after a software upgrade)

Issue #643, opened 2022-06-03
password reset link is broken in apex domain


donrestarone:
when the user tries to reset their password in the apex domain `example.com` the password reset link is incorrect (it points at `root.example.com` which results in a password reset token not valid error)

PR #670, opened 2022-06-07
[bug fix] password reset link is broken in apex domain

donrestarone:
addresses: https://github.com/restarone/violet_rails/issues/643

# acceptance criteria

1. password reset workflow works as expected in subdomain (`foo.example.com`)
2. password reset works as expected in apex domain (`example.com`)


donrestarone:
@alis-khadka since we dont have an email service setup for heroku-- 

I will promote this to staging as soon as I smoke test the app on heroku (if the app boots we are safe to promote)

alis-khadka:
> 2\. password reset works as expected in apex domain (`example.com`)

@donrestarone 
For apex domain, since the name is public (Apartment::Tenant.current), the url will be **public.example.com** instead of **root.example.com**

PR #678, opened 2022-06-07
[feature] add code editor to ExternalApiConnection

donrestarone:
# acceptance criteria

1. does not mutate/corrupt existing data when the form is rendered and submitted with no changes
2. renders existing data correctly


donrestarone:
@Pralish I tested this on Heroku and looks good, good to be promoted to staging

Issue #669, opened 2022-06-06
Logging not working on EC2


donrestarone:

donrestarone:
we added the 12 factor gem here: https://github.com/restarone/violet_rails/commit/6a1b891dc4204b2c44f56bf905978a14f8737d15

so the fact that we are not writing logs to files in production is expected: https://github.com/heroku/rails_12factor#rails-4-serve-static-assets


error reporting is working, but we will need to override the log behavior for it to actually write to disk.

PR #685, opened 2022-06-08
remove 12 factor gem to allow logging to disk

donrestarone:
addresses: https://github.com/restarone/violet_rails/issues/669

remove 12 factor gem, we have to make sure that we can still deploy to heroku without it-- the reason for removal is that it treats logs as event streams (we dont have an out-of-the-box solution for logging to stdout) and some clients need to write logs to disk


donrestarone:
Heroku deployment works without the 12 factor gem :+1:

Issue #655, opened 2022-06-04
add permission can_manage_files to decouple can_manage_web from file system access


donrestarone:
currently, only users with `can_manage_web: true` can access `Files`-- we want to add a new permission, can_manage_files that gate this access.

Issue #648, opened 2022-06-03
[Violet Rails Core]: Change scoping of api_clients and redirection after api_resource is deleted.


alis-khadka:
![image](https://user-images.githubusercontent.com/25191509/171900369-349076d8-21f9-4c28-8fba-e316c03fe41c.png)
In the above image, the url gives the indication that the displayed api_clients are under a specific api_namespace. But, we are displaying all the api_clients in this page.
```ruby
# GET /api_clients or /api_clients.json
  def index
    @api_clients = ApiClient.all
  end
```

And the redirect is wrong too (we should remove the view that is rendered on the redirect after the user deletes an API resource).

donrestarone:
https://api.rubyonrails.org/classes/ActionController/Redirecting.html#method-i-redirect_back

![Screenshot from 2022-06-06 20-50-01](https://user-images.githubusercontent.com/35935196/172272961-5228976f-512b-49bf-bc13-5d4e5b8421a2.png)

PR #686, opened 2022-06-08
[bug fix] remove 12 factor gem that was preventing logging on ec2, [feature] Change scoping of api clients and redirection after api resource is deleted, [bug fix] fix data discovery issue for ExternalApiClient in rake tasks (public schema is not being read), [feature] add permission can_manage_files to decouple can_manage_web from file system access

donrestarone:
## [bug fix] re-enable logging in EC2
addresses: https://github.com/restarone/violet_rails/issues/669

remove 12 factor gem, we have to make sure that we can still deploy to heroku without it-- the reason for removal is that it treats logs as event streams (we dont have an out-of-the-box solution for logging to stdout) and some clients need to write logs to disk

## [feature] Change scoping of api clients and redirection after api resource is deleted
addresses: https://github.com/restarone/violet_rails/issues/648

### acceptance criteria

1. lists API clients scoped to an API resource
2. after an API resource is destroyed it redirects to the correct place (not the API Resource #index, which should be removed)

## [feature] add permission can_manage_files to decouple can_manage_web from file system access

addresses: https://github.com/restarone/violet_rails/issues/655 

# acceptance criteria: 

1. after deployment file system access should be blocked for everyone
2. only users with permission can_manage_files should be able to access any view/action in `Files`
4. only users with permission can_manage_web can access `Layouts`, `Pages`, `Snippets`


donrestarone:
this fixed logging on `restarone.solutions` good to be promoted to production

donrestarone:
@alis-khadka this release candidate is ready for deployment! After monitoring overnight I will deploy tomorrow morning (5AM EST)

PR #687, opened 2022-06-08
[feature] Change scoping of api clients and redirection after api resource is deleted

donrestarone:
addresses: https://github.com/restarone/violet_rails/issues/648

# acceptance criteria

1. lists API clients scoped to an API resource
2. after an API resource is destroyed it redirects to the correct place (not the API Resource #index, which should be removed)


donrestarone:
testing is passing on Heroku @alis-khadka marking this as safe to promote to production

Issue #570, opened 2022-05-14
sanitize method/model definitions coming into Violet Rails of things like: Subdomain, Apartment, Tenant, switch, SomeModel.destroy_all, SomeModel.update_all, global_admin, can_manage_users


alis-khadka:
We are using 'eval(#custom code)' in our code.
We need to implement a way to scrub method/model definitions coming into Violet Rails of things like: Subdomain, Apartment, Tenant, switch, Subdomain.destroy_all, Subdomain.update_all, global_admin, can_manage_users, -- so we can open this upto platform developers on subdomains instead of whole systems (like how we have it now, because of this vulnerability).

donrestarone:
implement this as a before_save validation

donrestarone:
# external_api_clients => model_definition

https://github.com/restarone/violet_rails/blob/master/app/models/api_action.rb#L64-L67

PR #705, opened 2022-06-09
[feature] sanitize method/model definitions coming into Violet Rails of things like: Subdomain, Apartment, Tenant, switch, SomeModel.destroy_all, SomeModel.update_all, global_admin, can_manage_users

donrestarone:
addresses: https://github.com/restarone/violet_rails/issues/570

# acceptance criteria:

1. after deployment existing model definitions for ExternalAPIClient are not broken
2. user cannot perform dynamic programming in model definitions
3. user cannot escalate privileges in model definition
4. user cannot access other subdomains in model definition

### DIFF
--- a/app/models/api_action.rb
+++ b/app/models/api_action.rb
@@ -19,6 +19,8 @@ class ApiAction < ApplicationRecord
   has_rich_text :custom_message
 
   validates :http_method, inclusion: { in: ApiAction::HTTP_METHODS}, allow_blank: true
+  
+  validates :payload_mapping, safe_executable: true
 
   def self.children
     ['new_api_actions', 'create_api_actions', 'show_api_actions', 'update_api_actions', 'destroy_api_actions', 'error_api_actions']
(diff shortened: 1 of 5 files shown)


donrestarone:
@Pralish your feature is on its way to Heroku, lets use this as the base branch for this feature moving forward!

donrestarone:
@Pralish I am able to add malicious code to HTTP action in API Action:
Screen Shot 2022-06-09 at 6 50 08 AM

Pralish:
@donrestarone , Sorry about that. Here's the fix.

https://github.com/restarone/violet_rails/pull/706/

donrestarone:
@Pralish no worries! This is why we test manually as well-- deploying the latest change to Heroku now

donrestarone:
@Pralish privilege escalation is still possible: 

Screen Shot 2022-06-09 at 8 40 56 AM

PS: we should add test cases that explicitly block actions like this. Any privilege escalation (eg: changing any column that says `can_do_something: ` should be illegal)

donrestarone:
@Pralish thanks for the quick patch. This is working as expected.
Screen Shot 2022-06-09 at 12 07 30 PM
Screen Shot 2022-06-09 at 12 07 06 PM

I expect this to be part of a release candidate later today.

Issue #653, opened 2022-06-04
add subject to send_email API action, make subject / body dynamic, expose API entities in context


donrestarone:
Currently the email address and body are hard coded. The goal of this task is to add a subject as well and make both the subject and body dynamic so we can use Ruby-style string interpolation using `#{}`

The API Namespace/Resource should be exposed in this context as well, so we can do something like:

subject: `"#{api_resource.properties[:name]} welcome!"`
body `"We can help you with #{api_resource.properties[:problem]} !"`

![Screenshot from 2022-06-04 09-40-04](https://user-images.githubusercontent.com/35935196/172003251-2c688c05-b9b5-4b57-817e-5183d42317b5.png)

Issue #652, opened 2022-06-04
add dynamic string to HTTP action request URL


donrestarone:
Currently the path has to be hard coded, we want to use Ruby-style string interpolation using `#{}`

The API Namespace/Resource should be exposed in this context as well, so we can do something like: 

`www.example.com/#{api_resource.api_namespace.metadata["my_custom_attribute"]}`

https://github.com/restarone/violet_rails/wiki/API:-Entities,-Form-Rendering,-Interfaces-and-Actions#example-discord-bot

PR #708, opened 2022-06-09
Dynamic attributes with access to api_resource

Pralish:
https://github.com/restarone/violet_rails/issues/653

https://github.com/restarone/violet_rails/issues/652


donrestarone:
@Pralish can you separate this out to its own branch/PR?

PR #714, opened 2022-06-10
[feature] add subject to send_email API action, make subject / body dynamic, expose API entities in context, [feature] add dynamic string to HTTP action request URL

donrestarone:
[add subject to send_email API action, make subject / body dynamic, expose API entities in context](https://github.com/restarone/violet_rails/issues/653)

[add dynamic string to HTTP action request URL](https://github.com/restarone/violet_rails/issues/652)

# acceptance criteria: 

1. Existing HTTP actions are functional :heavy_check_mark: 
2. HTTP action allows string interpolation in URL :heavy_check_mark: 
3. Existing send email actions are functional
4. ability to use string interpolation in email, subject and body

### DIFF
--- a/app/controllers/comfy/admin/api_namespaces_controller.rb
+++ b/app/controllers/comfy/admin/api_namespaces_controller.rb
@@ -130,7 +130,7 @@ class Comfy::Admin::ApiNamespacesController < Comfy::Admin::Cms::BaseController
 
     # Only allow a list of trusted parameters through.
     def api_namespace_params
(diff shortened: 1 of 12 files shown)


donrestarone:
@Pralish I was able to validate half the features in your pull request on Heroku. For the other half (sending emails) we are testing on staging!

donrestarone:
@Pralish there's some interesting behavior happening for existing web requests / emails (on `support.restarone.solutions`) we see the following behavior: 

Screenshot from 2022-06-09 22-35-50

the system sent some empty emails and the email subject is now "send email"

Image from iOS


We will need to address these issues with existing web request / email actions before moving forward

Pralish:
@donrestarone , This issue has been addressed. The email subject will default to previous format if empty. Is that okay?

donrestarone:
@Pralish yup! When in doubt, we should stick to Violet Rails legacy

PR #716, opened 2022-06-10
Pralish dynamic attributes

Pralish:


donrestarone:
@Pralish emails are being sent as expected, but API actions related to HTTP requests are broken: 
Screen Shot 2022-06-10 at 8 15 45 AM

Issue #457, opened 2022-04-23
pass current_user / Ahoy::Visit to API/Data pipeline 


donrestarone:
`current_visit`, `current_user` (if any)

controller: app/controllers/resource_controller.rb

be able to use those the data in those objects in dynamic sections 

<img width="865" alt="Screen Shot 2022-06-13 at 8 46 33 AM" src="https://user-images.githubusercontent.com/35935196/173356962-a2f4a30d-54cb-4913-abfc-f972ddae7b7b.png">

PR #717, opened 2022-06-10
[feature] add subject to send_email API action, make subject / body dynamic, expose API entities in context, [feature] add dynamic string to HTTP action request URL

donrestarone:
[add subject to send_email API action, make subject / body dynamic, expose API entities in context](https://github.com/restarone/violet_rails/issues/653)

[add dynamic string to HTTP action request URL](https://github.com/restarone/violet_rails/issues/652)

# acceptance criteria: 

1. Existing HTTP actions are functional :heavy_check_mark: 
2. HTTP action allows string interpolation in URL :heavy_check_mark: 
3. Existing send email actions are functional
4. ability to use string interpolation in email, subject and body

### DIFF
--- a/app/controllers/comfy/admin/api_namespaces_controller.rb
+++ b/app/controllers/comfy/admin/api_namespaces_controller.rb
@@ -130,7 +130,7 @@ class Comfy::Admin::ApiNamespacesController < Comfy::Admin::Cms::BaseController
 
     # Only allow a list of trusted parameters through.
     def api_namespace_params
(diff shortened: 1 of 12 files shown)


donrestarone:
@Pralish emails are being sent as expected, but API actions related to HTTP requests are broken: 
Screen Shot 2022-06-10 at 8 15 45 AM

donrestarone:
[see here](https://github.com/restarone/violet_rails/issues/457) for more context about @alis-khadka

restarone-solutions:
@donrestarone, Can you pass me the payload_mapping for this api_action?

is it something like: self.properties["name"] ? 

If so, it should be self.api_resource.properties["name"] now.

previously self in this context was api_resource [I had changed the default behaviour]. But for consistency over all string interpolation, self will be the instance of model where dynamic string is defined [like it's suppose to be]

For example in this case, payload mapping is defined in api_action so the context of self will be api_action

PR #718, opened 2022-06-10
[Bug fix] last admin should not be able to delete themselves

donrestarone:
Addresses #313


donrestarone:
@Ayon95 this is on its way to Heroku!

donrestarone:
@Ayon95 looks good on Heroku. If there isn't another admin who can manage users, the delete button is not shown in the UI:
Screenshot from 2022-06-10 12-27-24
Screenshot from 2022-06-10 12-27-20


I will include it in the staging release later today, thanks!

PR #719, opened 2022-06-10
[Bug fix] last admin should not be able to delete themselves

donrestarone:
Addresses #313 

# acceptance criteria

1. delete button not shown if the user is the last admin who can manage users
2. admin cannot be deleted from the web console if they are the only one who can manage users


donrestarone:
@Ayon95 your change is on its way to `restarone.solutions` for `staging` confidence

Issue #105, opened 2021-05-21
add turbo support in CMS  / Forum


donrestarone:

donrestarone:
https://www.honeybadger.io/blog/hb-turbolinks-to-turbo/


https://github.com/hotwired/turbo-rails/blob/main/UPGRADING.md


https://hotwired.dev/#screencast


https://turbo.hotwired.dev/handbook/introduction

PR #751, opened 2022-06-14
[feature] hotwire/turbo support

donrestarone:
addresses: https://github.com/restarone/violet_rails/issues/105

Turbo works as expected in Web CMS pages. 
Screenshot from 2022-06-10 19-42-06

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -105,3 +105,5 @@ gem "exception_notification", "~> 4.5"
 
 
 gem "turnout", "~> 2.5"
+
+gem "turbo-rails", "~> 1.1"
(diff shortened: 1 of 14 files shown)


donrestarone:
update: since this is a big change, we want to run this on at least two of our systems before shipping to clients. On `restarone.solutions` the frontend is working as expected. But on `restarone.com` I noticed that the navbar animations are broken. 

this is a small issue and doesn't warrant an immediate roll-back. 

Lets keep monitoring turbo/hotwire for a while before deciding what to do next. 


But in the meantime, I set this as unsafe since we should **never ship platform upgrades that require client code changes to ensure expected behavior**

Issue #694, opened 2022-06-08
add tests for ExternalApiClientsController


donrestarone:
test/controllers/admin/comfy/external_api_clients_controller_test.rb

PR #757, opened 2022-06-14
[Violet Rails Core]: add tests for ExternalApiClientsController

alis-khadka:
Addressed https://github.com/restarone/violet_rails/issues/694

### DIFF
--- a/test/controllers/admin/comfy/external_api_clients_controller_test.rb
+++ b/test/controllers/admin/comfy/external_api_clients_controller_test.rb
@@ -103,17 +103,27 @@ class Comfy::Admin::ExternalApiClientsControllerTest < ActionDispatch::Integrati
     assert_response :success
   end
 
-  test "#create: should allow when permissioned user is signed-in" do
+  test "#create: should allow when permissioned user is signed-in and sets only the permitted atrributes" do
     payload = {
       external_api_client: {
         api_namespace_id: @api_namespace.id,
-        slug: 'create-test-api',
         label: 'Create Test API',
         enabled: true,
         metadata: {
                     "api_key": "x-api-key-foo",
-                    "bearer_token": "foo"
-                  },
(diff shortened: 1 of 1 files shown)


test/controllers/admin/comfy/external_api_clients_controller_test.rb
```diff
@@ -38,27 +75,402 @@ class Comfy::Admin::ExternalApiClientsControllerTest < ActionDispatch::Integrati
...
+    sign_in(@user)
+
+    get new_api_namespace_external_api_client_path(api_namespace_id: @api_namespace.id)
+
+    assert_response :success
+  end
+
+  test "#create: should allow when permissioned user is signed-in" do
+    payload = {
+      external_api_client: {
```
donrestarone, line 108: @alis-khadka can you add some assertions to this test that validates passed parameters are saved correctly on the DB (eg: metadata hash is saved the with the same keys/values)

test/controllers/admin/comfy/external_api_clients_controller_test.rb
```diff
@@ -38,27 +75,402 @@ class Comfy::Admin::ExternalApiClientsControllerTest < ActionDispatch::Integrati
...
+  test "#edit: allows if permissioned user is signed in" do
+    @user.update(can_manage_api: true)
+    sign_in(@user)
+
+    get edit_api_namespace_external_api_client_url(api_namespace_id: @api_namespace.id, id: @external_api_client.id)
+
+    assert_response :success
+  end
+
+  test "#update: should allow when permissioned user is signed-in" do
```
donrestarone, line 240: @alis-khadka can you add some assertions to this test that validates passed parameters are saved correctly on the DB (eg: metadata hash is saved the with the same keys/values, model_definition is correct etc)

Issue #692, opened 2022-06-08
fix data discovery issue in rake tasks (public schema is not accessed)


donrestarone:
I am exploring a hotfix here: https://github.com/restarone/violet_rails/pull/686/files#diff-bff8182f3a99c39531f1db7b69790c3b38596cb871726bd02088776c73c289be

Ayon95:
Can you provide some more context? What's the expected behavior? Also, you mentioned a hotfix. Did it work?

PR #758, opened 2022-06-14
fix(): include public subdomain when running rake tasks

Ayon95:
Addresses #692

### DIFF
--- a/app/models/subdomain.rb
+++ b/app/models/subdomain.rb
@@ -217,4 +217,9 @@ class Subdomain < ApplicationRecord
   def create_tenant
     Apartment::Tenant.create(self.name)
   end
+
+  def self.add_to_subdomains(subdomain_name)
+    subdomain = Subdomain.new(name: subdomain_name)
+    Subdomain.all.to_a.push(subdomain)
+  end
 end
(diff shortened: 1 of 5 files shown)


app/models/subdomain.rb
```diff
@@ -217,4 +217,9 @@ def self.bootstrap_via_comfy(name, hostname)
   def create_tenant
     Apartment::Tenant.create(self.name)
   end
+
+  def self.add_to_subdomains(subdomain_name)
```
donrestarone, line 221: @Ayon95 can you rename this to `self.all_with_public_schema` and don't take an argument.  
We don't need one because the `public` schema is special in the sense that it is always present in a Postgres DB (unless explicitly dropped)

Ayon95: All right, I'll make those changes.

lib/tasks/report.rake
```diff
@@ -1,9 +1,10 @@
 namespace :report do
   desc "subdomain analytics and report tasks"
-  
+
   task :send_analytics_report => [:environment] do 
     p "starting send_analytics_report task  @ #{Time.now}"
-    Subdomain.where.not(analytics_report_frequency: Subdomain::REPORT_FREQUENCY_MAPPING[:never]).each do |subdomain|
+    subdomains = Subdomain.all_with_public_schema
```
donrestarone, line 6: @Ayon95 nice! Before we were working with an activerecord collection, but with your change its just a simple array (because we shovel on an initialized subdomain) so we are filtering out any subdomain that is not reporting analytics

PR #762, opened 2022-06-15
fix data discovery issue in rake tasks (public schema is not accessed) 

donrestarone:
Addresses #692


donrestarone:
@Ayon95 this is symbolically and ceremonially going to Heroku. 

The real test will be on `restarone.solutions`

PR #765, opened 2022-06-15
fix data discovery issue in rake tasks (public schema is not accessed) 

donrestarone:
Addresses #692


donrestarone:
@Ayon95 your change is on its way to `restarone.solutions`

PR #766, opened 2022-06-15
allow multiple recipients for emails with select2

### LINKED ISSUE #149: select2 is not working in comfy/admin/cms area
we need to make sure select2 works on the email composition view

Screenshot from 2022-05-12 07-23-46

donrestarone:
resolves #149 

- connected select2 to recipients field
- previous `EMailer.ship` attempted to send separate email for each recipient, only the last recipient was being sent to
- EMailer now correctly fills the recipients array


donrestarone:
@yashkir this is on its way to Heroku!

PR #772, opened 2022-06-17
[Violet Rails Core]: Pass current user and current visit to api data pipeline

alis-khadka:
Addresses: https://github.com/restarone/violet_rails/issues/457


app/models/current.rb
```diff
@@ -0,0 +1,4 @@
+class Current < ActiveSupport::CurrentAttributes
```
donrestarone, line 1: @alis-khadka nice! We should use this singleton to replace calls to `Subdomain.current` that cost us a DB round trip

test/controllers/resource_controller_test.rb
```diff
@@ -191,4 +191,40 @@ class ResourceControllerTest < ActionDispatch::IntegrationTest
       post api_namespace_resource_index_url(api_namespace_id: api_namespace.id), params: payload
     end
   end
+
+  test 'should allow #create with the current_user & current_visit being available in the ' do
```
donrestarone, line 195: 'should allow #create with the current_user & current_visit being available in the send_web_request action'

test/controllers/resource_controller_test.rb
```diff
@@ -191,4 +191,40 @@ class ResourceControllerTest < ActionDispatch::IntegrationTest
...
+    @api_namespace.api_form.update(properties: { 'name': {'label': 'name', 'placeholder': 'Test', 'type_validation': 'tel'}})
+
+    api_action = api_actions(:one)
+    api_action.update!(
+                      type: 'CreateApiAction',
+                      bearer_token: 'test-token',
+                      position: 0,
+                      action_type: 'send_web_request',
+                      request_url: "http://www.example.com/success",
```
donrestarone, line 204: @alis-khadka I sort of get what this is testing, but we should have a sync about this test to make sure its not circular. 

not urgent, Happy to discuss during our weekly staff eng sync

Issue #783, opened 2022-06-20
nikean.orgpasswords#create (ActionView::MissingTemplate) "Missing template users/passwords/create, users/passwords


donrestarone:
``` txt
An ActionView::MissingTemplate occurred in passwords#create:

  Missing template users/passwords/create, users/passwords/create, devise/passwords/create, devise/create, application/create with {:locale=>[:en], :formats=>[:html], :variants=>[], :handlers=>[:raw, :erb, :html, :builder, :ruby, :jbuilder, :haml]}. Searched in:
  * "/var/www/violet/releases/20220619182518/app/views"
  * "/var/www/violet/shared/bundle/ruby/2.6.0/gems/ember-cli-rails-0.10.0/app/views"
  * "/var/www/violet/shared/bundle/ruby/2.6.0/gems/graphiql-rails-1.8.0/app/views"
  * "/var/www/violet/shared/bundle/ruby/2.6.0/gems/devise_invitable-2.0.3/app/views"
  * "/var/www/violet/shared/bundle/ruby/2.6.0/bundler/gems/simple_discussion-ccc76a485221/app/views"
  * "/var/www/violet/shared/bundle/ruby/2.6.0/bundler/gems/comfy-blog-767f3848165a/app/views"
  * "/var/www/violet/shared/bundle/ruby/2.6.0/bundler/gems/comfortable-mexican-sofa-456fb8507e96/app/views"
  * "/var/www/violet/shared/bundle/ruby/2.6.0/gems/devise-4.7.3/app/views"
  * "/var/www/violet/shared/bundle/ruby/2.6.0/gems/actiontext-6.1.5/app/views"
  * "/var/www/violet/shared/bundle/ruby/2.6.0/gems/actionmailbox-6.1.5/app/views"
```

Pralish:
## RCA

I have narrowed the issue down to this line on devise gem.

https://github.com/heartcombo/devise/blob/f8d1ea90bc328012f178b8a6616a89b73f2546a4/lib/devise/controllers/helpers.rb#L264


### Issue:

The request format is turbo_stream. It should have been html. Not sure when it started happening. This was suppose to happen after rails 7. But devise didn't recognize turbo_stream as navigational format.

I found there was a open PR to fix this on devise:
https://github.com/heartcombo/devise/pull/5340

**Actual**
[...]

PR #791, opened 2022-06-21
[bug fix] fix broken password reset

donrestarone:
resolves: https://github.com/restarone/violet_rails/issues/783

### DIFF
--- a/config/initializers/devise.rb
+++ b/config/initializers/devise.rb
@@ -312,7 +312,7 @@ Devise.setup do |config|
   # should add them to the navigational formats lists.
   #
   # The "*/*" below is required to match Internet Explorer requests.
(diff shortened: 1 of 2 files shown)


donrestarone:
@Pralish I couldn't test this end2end on Heroku since we don't have email setup there yet. So this is on its way to `staging` for alpha testing

donrestarone:
@Pralish the original bug is fixed, but I believe there are some other issues with our Devise integration. 

for example, `email confirmation resend` used to show some kind of flash message when you submitted this form: 
Screenshot from 2022-06-21 20-30-03

its not urgent, but when you get a chance can you create a github issue and triage?

donrestarone:
with that, merging to master -- thanks for your contribution @Pralish !

PR #793, opened 2022-06-22
remove model filter from test covergae

Pralish:


donrestarone:
Screen Shot 2022-06-23 at 7 26 20 AM

Issue #750, opened 2022-06-14
add matrix CI for Ruby 3 build


donrestarone:

donrestarone:
similar to: https://github.com/restarone/violet_rails/issues/660

donrestarone:
as part of this task, we should rename the Ruby Build action from : 

``` txt
build (14.x)
build (10.x)

```
``` txt
Violet Rails Build (Node.js 14.x) 
Violet Rails Build (Node.js 10.x) 
```

the new matrix action should be named: 

``` txt
Violet Rails Build (Ruby 2.6.6) 
Violet Rails Build (Ruby 3.0.0) 
```

so putting it all together we should have the following matrix: 

``` txt
Violet Rails Build (Node.js 14.x) 
Violet Rails Build (Node.js 10.x) 
Violet Rails Build (Ruby 2.6.6) 
Violet Rails Build (Ruby 3.0.0) 
```

PR #796, opened 2022-06-23
[Violet Rails Core] add matrix CI for Ruby 3 build

alis-khadka:
Addresses: https://github.com/restarone/violet_rails/issues/750

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -1,7 +1,11 @@
 source 'https://rubygems.org'
 git_source(:github) { |repo| "https://github.com/#{repo}.git" }
 
-ruby '2.6.6'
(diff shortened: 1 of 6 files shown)


Gemfile
```diff
@@ -1,7 +1,11 @@
 source 'https://rubygems.org'
 git_source(:github) { |repo| "https://github.com/#{repo}.git" }
 
-ruby '2.6.6'
+if ENV['RUBY_BUILD'] == '3.0.0'
```
donrestarone, line 4: @alis-khadka nice sensible default! we are using 2.6.6 in `production`

Issue #384, opened 2022-04-16
when feature is un-enabled remove buttons related to it


donrestarone:
<img width="443" alt="Screen Shot 2022-04-16 at 10 47 45 AM" src="https://user-images.githubusercontent.com/35935196/163685663-72e0689f-326e-427b-9701-77f66e697825.png">

PR #806, opened 2022-06-25
revert ruby 3 docker dev, remove model filter from test coverage, when feature is un-enabled remove buttons related to it

donrestarone:
Deploy not necessary (but will do once RC is packaged)

## change log

1. downgrade to ruby 2.6.6 for dev, since we run it in production its important we catch issues early on.  @donrestarone 
2. remove model filter from test coverage @Pralish 
3. https://github.com/restarone/violet_rails/issues/384 by @Gloria-kuang

### DIFF
--- a/Dockerfile.dev
+++ b/Dockerfile.dev
@@ -1,4 +1,4 @@
-FROM ruby:3.0.0-alpine
+FROM ruby:2.6.6-alpine
 
 # Build ENV vars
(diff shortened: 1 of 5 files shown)


donrestarone:
@Pralish @Gloria-kuang your changes are on staging! LGTM so far, I'll keep monitoring and will deploy to production tomorrow.

PR #808, opened 2022-06-25
Attempt to fix autoload deprecation

donrestarone:


donrestarone:
to fix the autoload deprecation warning we have to remove the gem ember-cli-rails and remove ember altogether

PR #764, opened 2022-06-15
when feature is disabled remove buttons related to it

Gloria-kuang:
This PR addresses the following issue:
#384 

When features are unselected, the buttons related to them are removed:
Screen Shot 2022-06-15 at 3 15 33 PM

PR #809, opened 2022-06-25
[feature] when feature is un-enabled remove buttons related to it (#764)

donrestarone:
resolves: https://github.com/restarone/violet_rails/issues/384

* remove buttons when feature is disable

* added negative cases

* updated the test units of the features


donrestarone:
@Gloria-kuang your change is on its way to Heroku for testing!

Issue #815, opened 2022-06-26
save code coverage report as artifact


donrestarone:
https://docs.github.com/en/actions/using-workflows/storing-workflow-data-as-artifacts

PR #828, opened 2022-06-27
[CI] Save code coverage as artifact

alis-khadka:
Addresses: https://github.com/restarone/violet_rails/issues/815

### DIFF
--- a/.github/workflows/heroku-deploy.yml
+++ b/.github/workflows/heroku-deploy.yml
@@ -1,26 +1,26 @@
-name: Deploy to Heroku
+# name: Deploy to Heroku
 
-on:
-  push:
-    branches: [ master, rc, dev ]
-  pull_request:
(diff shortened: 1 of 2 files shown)


alis-khadka:
> @alis-khadka how do we view test coverage? It's usually done by just opening the index.html file in the coverage folder

If we click on one of the link of artifact, a zip folder _(coverage folder)_ is downloaded. Within the folder, we can view the _index.html_ file.

https://user-images.githubusercontent.com/25191509/175912039-18e8d52e-7bbd-4850-b50a-67869d399fa6.mp4

donrestarone:
@alis-khadka yeah the default 90 days should be good! 

Just wanted to make sure we don't incur unnecessary charges since CI is run all the time

donrestarone:
@alis-khadka no that's okay, let's go with this for now and see how it goes!

PR #821, opened 2022-06-26
[feature] export/import API Namespace with/without children in JSON format

donrestarone:
Addresses: https://github.com/restarone/violet_rails/issues/679

### **Demo Video**

https://user-images.githubusercontent.com/25191509/175818688-42dc5179-dce9-4b22-a657-888ead6b3490.mp4

### **Screenshots**
image
image

PR #832, opened 2022-06-27
Export/Import API Namespace:  UI responsiveness & typo correction

alis-khadka:
https://github.com/restarone/violet_rails/pull/821#issuecomment-1167285746


donrestarone:
@alis-khadka I think you need to merge the latest changes from `master` into this branch before you target it against your feature branch on this repo https://github.com/restarone/violet_rails/pull/821

PR #840, opened 2022-06-29
[Bug Fix]: forum post/thread creation errors are not showing

alis-khadka:
RCA:
This issue seems to have started after we introduced hotwire-turbo in violet-rails.
Turbo expects 303(redirection) for form submission and 4xx/5xx to show validation errors in form.
Previously, when creating/updating of post/thread fails, we were not specifying status_code for the response.

https://turbo.hotwired.dev/handbook/drive#redirecting-after-a-form-submission

https://user-images.githubusercontent.com/35935196/176429422-cd40c98f-9ff3-49fa-a5cd-a64d21fcb283.mp4


donrestarone:
@alis-khadka can you add a test that exercises this flow? 

I'm hoping it would fail without your patch in place (ie: missing flash/redirect etc)

Issue #661, opened 2022-06-04
ability to toggle trix editor attachments


donrestarone:
We want to be able to use trix rich text without the ability to attach files (and vice versa) 

https://github.com/basecamp/trix/issues/604

donrestarone:
<img width="1728" alt="Screen Shot 2022-06-27 at 8 34 06 AM" src="https://user-images.githubusercontent.com/35935196/175942933-ff1ace0c-4128-46ef-a4ab-27ad644e11a0.png">

PR #858, opened 2022-07-01
[feature] ability to toggle trix editor attachments

donrestarone:
# changes

## [feature] ability to toggle trix editor attachments

Addresses: https://github.com/restarone/violet_rails/issues/661

### **Demo Video**

https://user-images.githubusercontent.com/35935196/176429682-a19cc7b7-f692-4f18-8bc9-7f95947fc9b2.mp4

### challenges
Unchecked checkbox not passing false to params:
It took me some time to figure this out. In the end, I followed these reference for the workaround;

    https://github.com/rails/rails/issues/5937
    https://stackoverflow.com/questions/20379311/how-to-get-blank-checkboxes-to-pass-as-false-to-params/20380051#20380051

### DIFF
--- a/app/controllers/comfy/admin/api_namespaces_controller.rb
+++ b/app/controllers/comfy/admin/api_namespaces_controller.rb
@@ -163,7 +163,7 @@ class Comfy::Admin::ApiNamespacesController < Comfy::Admin::Cms::BaseController
                                             :requires_authentication,
                                             :namespace_type,
                                             :has_form,
-                                            non_primitive_properties_attributes: [:id, :label, :field_type, :content, :attachment, :_destroy],
+                                            non_primitive_properties_attributes: [:id, :label, :field_type, :content, :attachment, :allow_attachments, :_destroy],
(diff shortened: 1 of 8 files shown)


donrestarone:
@alis-khadka the required rule does not work for rich text. This will need to be addressed. Happy to keep the release candidate under monitoring if you are able to submit a quick patch against this branch


Screen Shot 2022-07-01 at 6 34 28 PM

alis-khadka:
> @alis-khadka the required rule does not work for rich text. This will need to be addressed. Happy to keep the release candidate under monitoring if you are able to submit a quick patch against this branch
> 
> Screen Shot 2022-07-01 at 6 34 28 PM

The _required_ option was not working when I tested on old commit _(before my change)_.
It turns out that _required_ attribute works for native elements only.
reference: https://github.com/basecamp/trix/issues/328

donrestarone:
@alis-khadka yes, I believe so. But we are trying to replace HTML text area usage with this. 

Can we implement this on the client (just like we have for HTML text inputs/Text areas) by using trix for validation?

Issue #836, opened 2022-06-28
Migrating from Data Pipeline V2 to V3


Pralish:
- [ ] deprecate the magic self.custom_attribute syntax in favour of plain and explicit ruby (eg: api_resource.properties["custom_attribute"])

Manually find replace the magic syntax. I have written a script to list api_actions that uses the magic syntax. We will have to run this on each deployment server.

```
Subdomain.all_with_public_schema.each do |sd|
  res = []
  api_actions = ApiAction.where.not(api_namespace_id: nil).select {|a| a.payload_mapping.to_s.include?('#{')}
  api_actions.each do |api_action|
    p [sd.name, api_action.id, api_action.api_namespace_id]
  end
end
```

donrestarone:
@Pralish given the way the system is built, API Actions will continue to fail if this change isn't made right? Once the new syntax is in place, can we simply re-run the failed actions? (the only effect of the upgrade being, API actions related to HTTP actions failing temporarily)

Pralish:
> @Pralish given the way the system is built, API Actions will continue to fail if this change isn't made right? Once the new syntax is in place, can we simply re-run the failed actions? (the only effect of the upgrade being, API actions related to HTTP actions failing temporarily)

@donrestarone , Yes we can simply re-run the failed actions now. Thank you reminding about the failed actions. I have pushed the fix.

donrestarone:
@Pralish I'm targeting this weekend to wrap up testing and do the upgrade, will you be around?

donrestarone:
@Pralish can you redo the script? I dont think its working as you expect: 

``` ruby
irb(main):009:0> a.class
=> Array
irb(main):010:0> a[0]
[...]

PR #866, opened 2022-07-03
[feature] Dynamic attributes in API Actions

donrestarone:
# Dynamic segments in API Actions

This change adds dynamic segments in Email and Web Request Actions in the API Namespace / API Actions sub-systems.

addresses: https://github.com/restarone/violet_rails/issues/653

addresses: https://github.com/restarone/violet_rails/issues/652

upgrade path: https://github.com/restarone/violet_rails/issues/836

# Feature Showcase

## Dynamic Emails
Given a simple form like this: 
Screenshot from 2022-07-03 14-27-25

an email like this is triggered: 
Screenshot from 2022-07-03 14-24-27

the system will generate the email message: 
Screenshot from 2022-07-03 14-25-07

## Dynamic HTTP requests
Now you have dynamic segments in the request URL
Screen Shot 2022-07-04 at 9 30 47 AM

## View evaluated work
Every API Action is tracked by Violet Rails, and you can see what the system did after it evaluated your code. 

Screen Shot 2022-07-04 at 9 39 22 AM

The above action  resulted in the following email: 

Screen Shot 2022-07-04 at 9 41 11 AM

The same evaluation rules apply to HTTP actions, so you can see exactly what was sent to external systems -- for example: 
Screen Shot 2022-07-04 at 2 20 45 PM

### DIFF
--- a/app/controllers/comfy/admin/api_namespaces_controller.rb
+++ b/app/controllers/comfy/admin/api_namespaces_controller.rb
@@ -156,7 +156,7 @@ class Comfy::Admin::ApiNamespacesController < Comfy::Admin::Cms::BaseController
 
     # Only allow a list of trusted parameters through.
     def api_namespace_params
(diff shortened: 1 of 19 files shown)


donrestarone:
Monitoring until EOD Monday

Issue #789, opened 2022-06-21
Can't define a custom success/fail message without setting up a redirect action


Ayon95:
If no redirect action is specified, custom success message is not shown after submitting a form. The default success message is shown.
![default_success_flash_message](https://user-images.githubusercontent.com/73725297/174883342-39a88073-3b0d-42dd-a37d-a0bde9b0be86.png)

PR #880, opened 2022-07-06
[bug fix] Cannot define a custom success/fail message without setting up a redirect action

donrestarone:
Addresses: https://github.com/restarone/violet_rails/issues/789


donrestarone:
@alis-khadka this is on its way to Heroku!

Issue #872, opened 2022-07-04
API action workflow email action text area overflow


donrestarone:
**Describe the bug**

Text does not show correctly on the *vertical* axis

<img width="1728" alt="Screen Shot 2022-07-04 at 2 27 10 PM" src="https://user-images.githubusercontent.com/35935196/177203903-0e03d5d7-dfbc-467d-865c-45b3c3b6bcdd.png">

**To Reproduce**
1. Go to an API Namespace (if you don't have one, create one by visiting API)
2. Click on Action Workflow
3. Create an Email action
4. notice that the text field does not expand as the user types

**Expected behavior**

rich text area should expand

Ayon95:
The overflow is not happening on my machine.
![Capture](https://user-images.githubusercontent.com/73725297/177672546-aa4301e0-eee4-43ae-a44d-252b583d3dc0.PNG)

Ayon95:
All right, I'll submit a PR with the fix later today.

PR #885, opened 2022-07-08

donrestarone:
Addresses #872 

Capture


donrestarone:
@Ayon95 this is on its way to Heroku for UAT!

Issue #873, opened 2022-07-04
allow dynamic segment in API redirect action


donrestarone:
**Is your feature request related to a problem? Please describe.**

currently we can only redirect to pages. We want the ability to do both (pages + dynamic URL string)
<img width="1544" alt="Screen Shot 2022-07-04 at 2 45 56 PM" src="https://user-images.githubusercontent.com/35935196/177205432-6697cbff-d4f7-464a-9330-fe3994c18eb5.png">

**Describe the solution you'd like**

 using a radio select, allow the user to

1.  redirect using CMS page (existing UI)
2.  redirect using dynamic string (using a text input field)

This way we should be able to access `api_resource` and be able to do something like:

{`api_resource.properties['boolean'] ? root_path : some_other_path `}

*this task should only be started after this PR is merged to master: https://github.com/restarone/violet_rails/pull/866*

PR #887, opened 2022-07-08
[feature] Allow dynamic segment in api redirect action 

donrestarone:
# Allow dynamic segment in api redirect action

This feature allows Ruby evaluation in the redirect action so you can do conditional redirects and other fanciness: 

Screen Shot 2022-07-08 at 11 14 21 AM

---
## How it works

The implicit return value of the evaluated statement should be a valid path in the application. As illustrated above, you can use an absolute path (like: `'/'`, or a Rails route helper like `dashboard_path`)

---

Addresses: https://github.com/restarone/violet_rails/issues/873

## Video demo

https://user-images.githubusercontent.com/35935196/178011977-82aab515-87b5-49a3-8591-f8ba3535d9eb.mp4

## The evaluated output of the dynamic_url segment  is saved in the lifecycle message
image

### DIFF
--- a/app/views/comfy/admin/api_actions/_form.html.haml
+++ b/app/views/comfy/admin/api_actions/_form.html.haml
@@ -6,7 +6,7 @@
     .form-group.col-12
       = label_tag "Action"
       = select_tag "api_namespace[#{type}_attributes][#{index}][action_type]", options_for_select(ApiAction.action_types.keys, resource.action_type), { class: "form-control form-control-sm", id: "action_type_field_#{type}_#{index}", onchange: "toggleApiActionFields('#{index}', '#{type}');"}
-      - if resource.action_type == 'send_email' || resource.action_type == 'send_web_request'
+      .extra-details{ style: "#{'display: none' unless (resource.send_email? || resource.send_web_request? || (resource.redirect? && resource.dynamic_url?))}"}
         .form-group.my-2
           .d-block
             %strong Top-Level Model Attributes: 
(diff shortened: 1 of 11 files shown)


donrestarone:
@alis-khadka this is on its way to Heroku for UAT

donrestarone:
@alis-khadka here's the UAT feedback: 

## 1. Add inline-documentation for the dynamic segment
When the user selects dynamic_url they should be shown the variables that can be used in this segment (eg: `api_resource`, `rails route helpers eg: root_path`)
Screen Shot 2022-07-08 at 11 00 47 AM


## 2. the lifecycle message should include the evaluated output of the dynamic_url segment:

Screen Shot 2022-07-08 at 10 59 33 AM

app/views/comfy/admin/api_actions/_form.html.haml
```diff
@@ -19,11 +19,20 @@
...
+              \\#{api_resource.properties['your_custom_attribute_here']}
+        .mt-2.redirect-field-guide{ style: "#{'display: none' unless (resource.redirect? && resource.dynamic_url?)}"}
+          %small.d-block
+            %strong
+              Available Variables:
+            api_resource, rails-route-helpers (example: root_path)
+          %small.d-block
+            Dynamic:
+            %code
+              \\#{ api_resource.properties['your_custom_attribute_here'] == 'true' ? rooth_path : some_other_path }
```
donrestarone, line 35: @alis-khadka  please replace `rooth_path` with `comfy_admin_cms_path`
Screenshot from 2022-07-10 13-52-58

test/controllers/resource_controller_test.rb
```diff
@@ -192,6 +192,51 @@ class ResourceControllerTest < ActionDispatch::IntegrationTest
...
+      actions_count = @api_namespace.create_api_actions.size
+      assert_difference "@api_namespace.executed_api_actions.count", actions_count do
+        post api_namespace_resource_index_url(api_namespace_id: @api_namespace.id, params: payload)
+        assert_response :redirect
+        assert_redirected_to api_namespaces_path
+      end
+    end
+  end
+
+  test 'should allow #create and should not redirect by evaluating redirect_url if redirect_type is not dynamic_url' do
```
donrestarone, line 217: @alis-khadka Could you explain this test case to me? I dont think I understand it

test/controllers/resource_controller_test.rb
```diff
@@ -192,6 +192,51 @@ class ResourceControllerTest < ActionDispatch::IntegrationTest
     end
   end
```
donrestarone, line 193: @alis-khadka could you add a test that verifies the-non dynamic case? Where we simply redirect to a CMS page 

See below for example: 
Screenshot from 2022-07-10 14-01-15

test/controllers/resource_controller_test.rb
alis-khadka: In this test, I am setting the redirect_url to: `"\#{ api_resource.properties['flag'] == 'true' ? dashboard_path : api_namespaces_path }"` but it is not a dynamic redirect-type. 
So, when the redirection is being done, it should redirect explicitly to `"\#{ api_resource.properties['flag'] == 'true' ? dashboard_path : api_namespaces_path }"` **without evaluating it**.

donrestarone:
## UAT result on Heroku

1. General regression ✔️ 
2. dynamic redirect ✔️

PR #889, opened 2022-07-09

donrestarone:
Addresses #872 

Capture

### DIFF
--- a/app/views/comfy/admin/api_actions/_form.html.haml
+++ b/app/views/comfy/admin/api_actions/_form.html.haml
@@ -41,7 +41,7 @@
 
       .form-group
         = label_tag 'Custom Message'
-        = rich_text_area_tag "api_namespace[#{type}_attributes][#{index}][custom_message]", resource.custom_message, { class: "form-control form-control-sm" }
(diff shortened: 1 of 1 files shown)


app/views/comfy/admin/api_actions/_form.html.haml
```diff
@@ -1,125 +1,125 @@
...
+        = label_tag 'Email Address'
+        = text_field_tag "api_namespace[#{type}_attributes][#{index}][email]", resource.email, { class: "form-control form-control-sm" }
+
+      .form-group
+        = label_tag 'Subject'
+        =  text_field_tag "api_namespace[#{type}_attributes][#{index}][email_subject]", resource.email_subject, { class: "form-control form-control-sm" }
+
+      .form-group
+        = label_tag 'Custom Message'
+        = rich_text_area_tag "api_namespace[#{type}_attributes][#{index}][custom_message]", resource.custom_message, { class: "form-control form-control-sm", style: "height: auto;" }
```
donrestarone, line 44: @Ayon95 come to think of it, I don't think we should merge this PR because this diff is far too big for just this change. Can you double check why this is the case? It's not typical to make a small change like this and get a huge diff like this.

PR #890, opened 2022-07-09

donrestarone:
Addresses #872 

Capture

### DIFF
--- a/app/views/comfy/admin/api_actions/_form.html.haml
+++ b/app/views/comfy/admin/api_actions/_form.html.haml
@@ -41,7 +41,7 @@
 
       .form-group
         = label_tag 'Custom Message'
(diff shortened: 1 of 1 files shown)


donrestarone:
@Ayon95 The diff [here](https://github.com/restarone/violet_rails/pull/889/files) looks like there's something wrong with your git or you are using different spaces (we use 2 spaces in the codebase) or a linter-- but anyway, here's what the diff should look like. 

Happy to help with the git issue if you need the support

Ayon95:
I think It could be because of a line-ending issue and how Windows and Unix-based operating systems treat line endings differently. In my Git config, `core.autocrlf` is set to `true`.

https://stackoverflow.com/questions/30752959/git-showing-entire-file-is-modified-instead-of-showing-modified-small-part-of-co
https://docs.github.com/en/get-started/getting-started-with-git/configuring-git-to-handle-line-endings
https://git-scm.com/book/en/v2/Customizing-Git-Git-Configuration

Capture1

PR #892, opened 2022-07-09
[feature] Pass current user and current visit to api data pipeline

donrestarone:
## ability to reference `current_user` and `current_visit` in API Actions
Addresses: https://github.com/restarone/violet_rails/issues/457
Screenshot from 2022-07-11 17-57-56
## example usage
Screenshot from 2022-07-11 17-58-10

### DIFF
--- a/app/models/api_action.rb
+++ b/app/models/api_action.rb
@@ -4,7 +4,7 @@ class ApiAction < ApplicationRecord
   include DynamicAttribute
 
   attr_encrypted :bearer_token
-  attr_dynamic :email, :email_subject, :custom_message, :payload_mapping, :request_url
+  attr_dynamic :email, :email_subject, :custom_message, :payload_mapping, :request_url, :custom_headers
 
   after_update :update_executed_actions_payload, if: Proc.new { api_namespace.present? && saved_change_to_payload_mapping? }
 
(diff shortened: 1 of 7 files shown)


app/models/api_action.rb
```diff
@@ -73,7 +74,7 @@ def redirect;end
   def serve_file;end
 
   def request_headers
-    headers = custom_headers.to_json.gsub('SECRET_BEARER_TOKEN', bearer_token)
+    headers = custom_headers_evaluated.gsub('SECRET_BEARER_TOKEN', bearer_token)
```
donrestarone, line 77: @alis-khadka does this allow dynamic headers here? (see below) -- if so can you add in-line documentation
Screenshot from 2022-07-09 16-21-59

donrestarone:
@alis-khadka can you attach some screenshots and a demo video? I believe in-line documentation for the variables in scope would be nice too!

app/models/api_action.rb
```diff
@@ -26,6 +26,7 @@ class ApiAction < ApplicationRecord
   validates :http_method, inclusion: { in: ApiAction::HTTP_METHODS}, allow_blank: true
   
   validates :payload_mapping, safe_executable: true
+  validates :custom_headers, safe_executable: true
```
Pralish, line 29: @alis-khadka , all dynamic attributes are safe by default. You don't need to validate it here. The validation for payload_mapping can also be removed.

donrestarone:
## :spiral_notepad: UAT on Heroku underway

Issue #767, opened 2022-06-15
[feature] ability to have an email signature + custom from in email


donrestarone:
define email signature + email from (name) in web settings 

<img width="1572" alt="Screen Shot 2022-07-09 at 12 54 25 PM" src="https://user-images.githubusercontent.com/35935196/178115452-55e9a931-45a6-45f0-83bc-e04c18be22eb.png">

and then use it in every email that is sent

<img width="460" alt="Screen Shot 2022-06-21 at 7 52 47 AM" src="https://user-images.githubusercontent.com/35935196/174792924-142c7f46-0b29-4a69-b47b-20a162c9bf34.png">
<img width="1410" alt="Screen Shot 2022-06-21 at 7 52 42 AM" src="https://user-images.githubusercontent.com/35935196/174792925-60d1568a-f14d-4f6e-8c37-df96bf0b4ad7.png">
<img width="606" alt="Screen Shot 2022-06-21 at 7 51 40 AM" src="https://user-images.githubusercontent.com/35935196/174792927-30ff2228-00cd-4145-b4ce-0d98fde06bc9.png">

donrestarone:
@Pralish feel free to pass this along to @Ayon95 if you're looking into other tasks this coming sprint

donrestarone:
# [Implementation Proposal] ability to have an email signature + custom from in email
This is my proposal for adding the feature to Violet Rails Core

---
## Persisting the name of the email'er
add `email_name` `string` column to `subdomains` table: https://github.com/restarone/violet_rails/blob/master/db/schema.rb#L450-L479
---
## Add name to outgoing email
reference: https://stackoverflow.com/questions/20811353/how-can-we-set-the-name-of-the-email-sender-in-rails-mailer
Add `from: 'Sender Name <sender@example.com>'` to https://github.com/restarone/violet_rails/blob/master/app/mailers/e_mailer.rb#L13
---
## Persisting the email signature of the email'er
1. add `email_signature` `text` column to `subdomains` table: https://github.com/restarone/violet_rails/blob/master/db/schema.rb#L450-L479

2. make it a rich text field by adding `has_rich_text : email_signature ` to the `Subdomain` model (eg: https://github.com/restarone/violet_rails/blob/master/app/models/message.rb#L7)3. 

3. add the rich text input to the `web_settings` form: https://github.com/restarone/violet_rails/blob/master/app/views/comfy/admin/web_settings/_form.haml4. 

[...]

PR #898, opened 2022-07-09
[feature] ability to have an email signature + name in email

donrestarone:
# [feature] ability to have an email signature + name in email

resolves: https://github.com/restarone/violet_rails/issues/767
## Set name and signature for Email under App Settings
Screenshot from 2022-07-09 15-58-10

image

## Name and signature are included in outgoing Email

image


donrestarone:
Monitoring until EOD Monday July 10th 2022

PR #918, opened 2022-07-12
[feature] Allow dynamic segment in API redirect action

donrestarone:
# Allow dynamic segment in api redirect action

This feature allows Ruby evaluation in the redirect action so you can do conditional redirects and other fanciness: 

Screen Shot 2022-07-08 at 11 14 21 AM

---
## How it works

The implicit return value of the evaluated statement should be a valid path in the application. As illustrated above, you can use an absolute path (like: `'/'`, or a Rails route helper like `dashboard_path`)

---

Addresses: https://github.com/restarone/violet_rails/issues/873

## Video demo

https://user-images.githubusercontent.com/35935196/178011977-82aab515-87b5-49a3-8591-f8ba3535d9eb.mp4

## The evaluated output of the dynamic_url segment  is saved in the lifecycle message
image

Co-authored-by: Prashant Khadka <prashant.khadka052@gmail.com>

### DIFF
--- a/.github/workflows/deploy.yml
+++ b/.github/workflows/deploy.yml
@@ -5,7 +5,7 @@ on:
   workflow_dispatch:
 jobs:
   build:
-    runs-on: ubuntu-18.04
+    runs-on: ubuntu-latest
     steps:
       - name: Choose Environment
         id: branch_check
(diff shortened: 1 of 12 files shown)


donrestarone:
## Staging confidence

1. General regression :heavy_check_mark: 
2. Performance :hourglass: 
3. Monitoring :hourglass:

Issue #900, opened 2022-07-09
(if any) attach current_visit and current_user to created API Resource 


donrestarone:
**Is your feature request related to a problem? Please describe.**

Currently we dont track the user/visit for when forms are submitted. 

**Describe the solution you'd like**

Track the user/visit for when forms are submitted and API resources are created

![Screenshot from 2022-07-09 16-24-00](https://user-images.githubusercontent.com/35935196/178121502-62e5ce5d-52cb-4406-826a-5cae197848f6.png)

PR #939, opened 2022-07-19
attach current_visit and current_user to created API Resource

Pralish:
Addresses: https://github.com/restarone/violet_rails/issues/900

### DIFF
--- a/app/controllers/concerns/api_actionable.rb
+++ b/app/controllers/concerns/api_actionable.rb
@@ -6,9 +6,10 @@ module ApiActionable
     before_action :check_for_redirect_action, only: [:create, :update, :show, :destroy]
     after_action :execute_api_actions, only: [:show, :create, :update, :destroy]
     before_action :check_for_serve_file_action, only: [:show, :create, :update, :destroy]
+    after_action :track_create_event, only: :create
     rescue_from StandardError, with: :handle_error
   end
-
+  
 
   def check_for_redirect_action
     @redirect_action = if @api_resource.present?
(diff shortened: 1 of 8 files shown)


app/controllers/concerns/api_actionable.rb
```diff
@@ -105,4 +106,11 @@ def load_api_actions_from_api_resource
     @redirect_action = @api_resource.send(api_action_name).where(action_type: 'redirect').reorder(:created_at).last if @redirect_action.present?
     @serve_file_action = @api_resource.send(api_action_name).where(action_type: 'serve_file').reorder(:created_at).last if @serve_file_action.present?
   end
+
+  def track_create_event
+    ahoy.track(
+      "api-resource-create",
+      { visit_id: current_visit.id, api_resource_id: @api_resource.id, user_id: current_user&.id }
```
donrestarone, line 113: @Pralish could be useful to tag the API Namespace ID here as well

app/views/comfy/admin/api_resources/show.html.haml
```diff
@@ -9,6 +9,15 @@
 = render 'comfy/admin/api_namespaces/editor', {api_resource: @api_resource, form_properties: @api_resource.api_namespace&.api_form&.properties&.symbolize_keys, read_only: true}
 = render 'comfy/admin/api_namespaces/init_editor', {read_only: true}
 
+%p
+  %b Current User:
```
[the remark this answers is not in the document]
Pralish: I made the change.

app/controllers/concerns/api_actionable.rb
Pralish: It's done. Thanks for the feedback

Issue #915, opened 2022-07-12
users are not redirected properly after sign in


donrestarone:
**Describe the bug**

Suppose there is a page behind a login (a private page), when the user visits such a page-- they are presented with the login. But when they provide their login credentials and hit submit, they are directed to the Violet Rails admin instead of the original page they were trying to access (the private page mentioned earlier). 

**To Reproduce**
1. sign out
2. click on a link that takes you to a private page (could be a CMS page or a private forum thread)
3. login is presented
4. sign in 
5. redirected to admin area 

**Expected behavior**

The after sign in path should be the page the user was originally trying to access before they got redirected to login first. 

## Discovery 

1. Afte sign in path is set here: https://github.com/restarone/violet_rails/blob/master/app/controllers/application_controller.rb

PR #957, opened 2022-07-22
[Bug Fix] User not redirected to previous location after sign in (#956)

donrestarone:
Addresses: https://github.com/restarone/violet_rails/issues/915
* fix(): saving the initial url for redirection

* feat(): refactor and add test case

* feat(): fix test case

### DIFF
--- a/test/controllers/users/sessions_controller_test.rb
+++ b/test/controllers/users/sessions_controller_test.rb
@@ -151,6 +151,18 @@ class Users::SessionsControllerTest < ActionDispatch::IntegrationTest
       # After signing-in, it should redirect back to the private cms page
       post user_session_url(subdomain: @restarone_subdomain.name), params: payload
       assert_redirected_to page.full_path
+
+      # It should redirect back to the root_url if the user cannot view restricted pages
+      @restarone_user.update!(can_view_restricted_pages: false)
+      sign_in(@restarone_user)
+      get comfy_cms_render_page_url(subdomain: @restarone_subdomain.name, cms_path: 'test-cms-page')
+      assert_redirected_to root_url(subdomain: @restarone_subdomain.name)
+
+      # It should render private cms-page successfully if the user can view restricted pages
+      @restarone_user.update!(can_view_restricted_pages: true)
+      sign_in(@restarone_user)
+      get comfy_cms_render_page_url(subdomain: @restarone_subdomain.name, cms_path: 'test-cms-page')
+      assert :success
     end
   end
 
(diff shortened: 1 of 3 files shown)


donrestarone:
@alis-khadka this is on its way to Heroku! 

## Acceptance criteria

1. When a non-signed in user navigates to a private forum thread `A`, they should be presented with a login screen. After login they should be redirected to the originally navigated to thread `A`
2. When a non-signed in user navigates to a private webpage `foo`, they should be presented with a login screen. After login they should be redirected to the originally navigated to page `foo`
3. When a non-signed in user navigates to a API Namespace `A`, they should be presented with a login screen. After login they should be redirected to the originally navigated to API Namespace `A`

test/controllers/users/sessions_controller_test.rb
```diff
@@ -124,4 +124,59 @@ class Users::SessionsControllerTest < ActionDispatch::IntegrationTest
     assert_template :new
     assert_match "Invalid Email or password.", flash[:alert]
   end
+
+  test 'redirects to private cms-page after sign-in successfully' do
+    Apartment::Tenant.switch @restarone_subdomain.name do
+      @restarone_user.update!(can_access_admin: false)
```
donrestarone, line 130: @alis-khadka double check this test case, feels like a false positive. The user doesn't have permission to access private pages

donrestarone:
@alis-khadka for completeness please add a test case that exercises the following: 

When a non-signed in user navigates to a API Namespace `A`, they should be presented with a login screen. After login they should be redirected to the originally navigated to API Namespace `A` :heavy_check_mark:

test/controllers/users/sessions_controller_test.rb
```diff
@@ -124,4 +124,111 @@ class Users::SessionsControllerTest < ActionDispatch::IntegrationTest
...
+
+      # After signing-in, it should redirect back to the api-namespace page
+      post user_session_url(subdomain: @restarone_subdomain.name), params: payload
+      assert_redirected_to api_namespace_url(subdomain: @restarone_subdomain.name, id: api_namespace.id)
+
+      # The response will be success if the user can manage api
+      @restarone_user.update!(can_manage_api: true)
+      sign_in(@restarone_user)
+      get api_namespace_url(subdomain: @restarone_subdomain.name, id: api_namespace.id)
+      assert :success
```
donrestarone, line 225: @alis-khadka shouldn't this be `assert_response :success` ? Im not sure what `assert :success` does, I don't see any docs on it eg: https://apidock.com/rails/ActionController/Assertions/ResponseAssertions/assert_response

If they are equivalent, happy to approve as is

Issue #924, opened 2022-07-14
cant delete user if they have forum post subscriptions


donrestarone:
**Describe the bug**
stack trace:

``` txt
An ActiveRecord::InvalidForeignKey occurred in users#destroy:

  PG::ForeignKeyViolation: ERROR:  update or delete on table "users" violates foreign key constraint "fk_rails_8268bd8830" on table "forum_subscriptions"
DETAIL:  Key (id)=(6) is still referenced from table "forum_subscriptions".

  app/controllers/comfy/admin/users_controller.rb:50:in `destroy'
```

PR #960, opened 2022-07-22
[Bug fix]: can't delete user if they have forum post subscriptions

alis-khadka:
Addresses: https://github.com/restarone/violet_rails/issues/924

**Demo**

https://user-images.githubusercontent.com/25191509/180488253-0c7f890a-9731-4cba-b8d0-0f2759145e90.mp4


donrestarone:
@alis-khadka Can we add a warning next to the delete button if the user has forum posts / attached analytics events? The warning could read something like: 

`Deleting the user will remove #{number_of_posts_or_threads} forum contributions and #{number_of_analytics_events} analytics events`

Issue #875, opened 2022-07-05
REST API and GQL does not return non primitive types (richtext and file)


donrestarone:
**Describe the bug**

The REST API and GQL should include non primitive type fields in the response. 

**To Reproduce**
Given an API Namespace like the following: 

![Screenshot from 2022-07-05 10-42-23](https://user-images.githubusercontent.com/35935196/177355570-309319be-3b06-4fdd-8b30-72e0c33f52f6.png)

the system does not include rich text strings in the REST API response: 

![Screenshot from 2022-07-05 10-42-11](https://user-images.githubusercontent.com/35935196/177357012-96811aad-041c-429a-974f-146c98e5fbb9.png)

you can see the Graph and REST interfaces here:

![Screenshot from 2022-07-05 10-43-58](https://user-images.githubusercontent.com/35935196/177357038-982d6ce9-c298-4f10-a413-60deb191ae60.png)

**Expected behavior**

For rich text the system should return a string under the respective attribute and for a file it should be a rails/activestorage generated URL.

Pralish:
**Upgrade to latest version of describe api**

- use response['data']['attributes'] instead of just response

PR #968, opened 2022-07-26
[Feature] return non primitive types in REST API and GraphQL response

Pralish:
Addresses: https://github.com/restarone/violet_rails/issues/875


test/controllers/graphql_controller_test.rb
```diff
@@ -64,6 +64,13 @@ class GraphqlControllerTest < ActionDispatch::IntegrationTest
     assert_equal ["data"], json_response.keys
   end
 
+  test "[if enabled] query public API Namespaces with nested apiResources with non-primitive properties" do
```
donrestarone, line 67: @Pralish can we add a more in depth test case that tests for: 
1. mime type & object URL 
2. rich text

Issue #949, opened 2022-07-21
API form success flash message randomly showing in admin UI


donrestarone:
**Describe the bug**

When navigating the Admin, on certain screens the user is presented with the API Form success flash message

<img width="1728" alt="Screen Shot 2022-07-20 at 8 38 21 PM" src="https://user-images.githubusercontent.com/35935196/180105765-acea8289-8974-4df7-9525-b8acd2882d1f.png">
<img width="1728" alt="Screen Shot 2022-07-20 at 8 38 18 PM" src="https://user-images.githubusercontent.com/35935196/180105768-d892a2be-bf63-4f70-a8f9-d82b75af36bb.png">
<img width="1728" alt="Screen Shot 2022-07-20 at 8 37 59 PM" src="https://user-images.githubusercontent.com/35935196/180105772-775b29c8-8b40-4486-9e18-c088856571ec.png">

Pralish:
## RCA

There seems to be no redirection happening when api form is submitted unless there is an explicit redirect api action defined.

Flash messages are only shown in next request i.e. after redirection. Since there's no redirection after form submission, the flash messages are showing up after any next request.


https://user-images.githubusercontent.com/50227291/180973705-7e050473-8247-4fcd-857c-7d452e75d27a.mov



#### FIX:

https://github.com/restarone/violet_rails/pull/969

donrestarone:
@Pralish looks like this issue has come back again:
<img width="878" alt="Screen Shot 2022-08-11 at 8 57 44 PM" src="https://user-images.githubusercontent.com/35935196/184266124-c5a5d556-8d2e-459f-bf21-748738a36e37.png">

donrestarone:
<img width="864" alt="Screen Shot 2022-08-15 at 8 46 15 AM" src="https://user-images.githubusercontent.com/35935196/184637613-689ce5a1-6f3d-4017-8273-506930c27c6d.png">

PR #969, opened 2022-07-26
[Bug Fix] Redirect back if no redirect action is defined

Pralish:
Addresses: https://github.com/restarone/violet_rails/issues/949

## Demo

https://user-images.githubusercontent.com/35935196/180993751-4c773627-5b41-4c93-a88e-8bcdd89ac2f0.mov


donrestarone:
@Pralish thanks for proactively picking this up! I was anticipating complaints coming in next week but I think with this patch we are ahead of the tickets ;)

PR #991, opened 2022-07-31
[Feature] return non primitive types in REST API and GraphQL response

donrestarone:
## [Feature] return non primitive types in REST API and GraphQL response
Addresses: https://github.com/restarone/violet_rails/issues/875

Screen Shot 2022-07-26 at 7 53 51 AM
Screen Shot 2022-07-26 at 7 54 26 AM

### DIFF
--- a/app/controllers/api/resource_controller.rb
+++ b/app/controllers/api/resource_controller.rb
@@ -18,7 +18,7 @@ class Api::ResourceController < Api::BaseController
   end
 
   def describe
-    render json: @api_namespace
+    render json: @api_namespace, include: :non_primitive_properties
   end
 
   def show
--- a/app/controllers/api/resource_controller.rb
+++ b/app/controllers/api/resource_controller.rb
@@ -22,7 +22,7 @@ class Api::ResourceController < Api::BaseController
   end
 
   def describe
-    render json: @api_namespace, include: :non_primitive_properties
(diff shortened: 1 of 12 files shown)


donrestarone:
## Acceptance criteria 

1. Create public API Namespace with primitive and non primitive properties ⌛ 
2. Create a few API Resources ⌛ 
3. Query via REST API => There should be URL's for files and text for rich text ⌛ 
4. Query via GraphQL => There should be URL's for files and text for rich text ⌛ 
5. GraphQL returns ActionText strings with embeds ⌛ 
6. REST API returns ActionText strings with embeds ⌛

donrestarone:
@Pralish 

# ❌ Blockers 

1. `/describe` does not include non primitive properties: 
Screen Shot 2022-07-31 at 10 03 30 AM
Screen Shot 2022-07-31 at 10 03 16 AM

2. Generated GQL documentation's slug isn't correct: 

notice the camelized slug
Screen Shot 2022-07-31 at 10 07 56 AM

when used results in a 404:

Screen Shot 2022-07-31 at 10 10 47 AM

but when we check the slug on the URL (in the app for the API Namespace) and use that it is correct: 
Screen Shot 2022-07-31 at 10 10 54 AM


3. Generated API documentation's apex domain is not correct if the system is using the public schema (eg: on Heroku)
Screen Shot 2022-07-31 at 10 12 31 AM

Issue #976, opened 2022-07-28
Nginx loadbalancer


donrestarone:
**Is your feature request related to a problem? Please describe.**

Currently we do not provide an out of the box load balancing solution for Violet Rails. Clients are left to use out-of-the-box options like, 

1. AWS load balancers
2. Heroku

Or if they prefer to roll their own (like we have in some instances), they are welcome to use the [NGINX recipes here](https://github.com/restarone/violet_rails/wiki/nginx-configurations)

**Describe the solution you'd like**

#### We need an out of the box solution!🥁 

**Research**

1. High level AWS Route 53 + Nginx setup: https://www.bogotobogo.com/DevOps/AWS/aws-Route53-DNS-Subdomain.php
2. Nginx on docker: https://codeburst.io/scaling-out-with-docker-and-nginx-8eda9fb1654c
3. Docker-compose with multiple containers: https://www.bogotobogo.com/DevOps/Docker/Docker-Compose-Nginx-Reverse-Proxy-Multiple-Containers.php
4. Nginx docs, proxying to many servers: https://docs.nginx.com/nginx/admin-guide/load-balancer/http-load-balancer/
5. Dockerized Rails app with Nginx: https://scotto.medium.com/2021-docker-ruby-3-rails-6-puma-nginx-postgres-d84c95f68637

Issue #972, opened 2022-07-26
SyncAttributeToApiNamespace Plugin : External API Connection that backfills attributes on API Namespace + API Resources


alis-khadka:
### SyncAttributeToApiNamespace Plugin
`
class SyncAttributeToApiNamespace
  def initialize(parameters)  
    @external_api_client = parameters[:external_api_client]
    @api_namespace = @external_api_client.api_namespace
    @api_form = @api_namespace.api_form
  end

  def start
    # Fetching provided inputs (attribute_name, default_value)
    attribute_name = @external_api_client.metadata["ATTRIBUTE_NAME"]
    placeholder_value = @external_api_client.metadata["PLACEHOLDER_VALUE"] || ''
    default_value = @external_api_client.metadata["DEFAULT_VALUE"] || ''

    raise 'The provided attribute is already defined in the ApiNamespace' if @api_namespace.properties.keys.include?(attribute_name)

    # Adding the new-attribute in ApiNamespace
    new_properties = @api_namespace.properties.merge(attribute_name => placeholder_value)
    @api_namespace.has_form = @api_namespace.api_form.present? ? '1' : '0'
    @api_namespace.update!(properties: new_properties)

    # Making the new-attribute non-renderable
    form_properties = @api_form.properties
    form_properties[attribute_name]['renderable'] = '0'
    @api_form.update!(properties: form_properties)

    # Adding the new-attribute to its ApiResources
    @api_namespace.api_resources.find_in_batches(batch_size: 500).each do |resources_batch|
      resources_batch.each do |api_resource|
        # Does not mutate the api-resource if it already has the provided new-attribute
        next if api_resource.properties.keys.include?(attribute_name)

        new_resource_properties = api_resource.properties.merge(attribute_name => default_value)
        api_resource.update!(properties: new_resource_properties)
      end
    end
  end
end
# at the end of the file we have to implicitly return the class 
SyncAttributeToApiNamespace
`ruby

The following metadata needs to be included for the plugin to work:
![image](https://user-images.githubusercontent.com/25191509/181057770-bd2adad8-c085-4f62-a6c3-35689e4362e7.png)

**ATTRIBUTE_NAME**: The name of the new attribute that needs to added.
**PLACEHOLDER_VALUE**: The value that will be shown as a placeholder. This will be added in the api-namespace properties' value.
**DEFAULT_VALUE**: The default value that will be added in the api-resources properties' value.

### **Output**
[...]

Issue #870, opened 2022-07-04
Create External API Connection that backfills attributes on API Namespace + API Resources


donrestarone:
**Is your feature request related to a problem? Please describe.**

If your API Namespace attributes change, there is no way to sync/backfill those new attributes with existing API Resources

**Describe the solution you'd like**

create an External API Connection similar to VacuumJob (https://github.com/restarone/violet_rails/blob/master/test/fixtures/external_api_clients.yml#L203) that backfills a given attribute. 

Running the External API Connection should: 

1. Add the attribute to the API Namespace
2. Backfill all API Resources with the attribute and its default value (if any)
3. not delete or mutate any other data
4. be safe to run multiple times ([idempotent](https://stackoverflow.com/questions/1077412/what-is-an-idempotent-operation))

PR #994, opened 2022-07-31
[Dev workflow] minimal config for nginx in docker + rails, [Plugin] Create External API Connection that backfills attributes on API Namespace + API Resources

donrestarone:
# Development changes only, no deploy required :heavy_check_mark: 

## [Dev workflow] minimal config for nginx in docker + rails

resolves: https://github.com/restarone/violet_rails/issues/976

### Use Nginx in Violet Rails dev! 

Screen Shot 2022-07-28 at 8 40 00 AM

## [Plugin] Create External API Connection that backfills attributes on API Namespace + API Resources

Addresses: https://github.com/restarone/violet_rails/issues/870
### Documentation: https://github.com/restarone/violet_rails/issues/972


donrestarone:
# [Plugin] Create External API Connection that backfills attributes on API Namespace + API Resources
## :x: Blockers
@alis-khadka I have merged your change to this No Deploy Required :tm:  branch, here are the merge blockers: 

### 1. `false` default value is not respected, the placeholder value is inserted into the default value instead. 
Given the following configuration: 
Screenshot from 2022-07-31 13-01-47
we expect the default value to be backfilled as `false`, but that is not the case: 
Screenshot from 2022-07-31 13-05-37
Screenshot from 2022-07-31 13-05-29

donrestarone:
@alis-khadka now the default value is being backfilled on the API Resources, but the same issue is happening on the parent namespace. We expect both values to be set to the same default value.
Screen Shot 2022-08-01 at 10 29 16 AM
Screen Shot 2022-08-01 at 10 29 10 AM

PR #1012, opened 2022-08-10
[Violet Rails Core]: Fire api actions from model

alis-khadka:
Addresses: https://github.com/restarone/violet_rails/issues/584


Pralish:
@alis-khadka, Found the solution to fix the failing test cases.

You need to do `Sidekiq::Testing.inline!` for async jobs to avoid the race conditions.

Ref: https://www.cloudbees.com/blog/know-your-sidekiq-testing-rights

Issue #980, opened 2022-07-29
Render API Resource file attachment if representable (eg: video/ image)


donrestarone:
**Is your feature request related to a problem? Please describe.**

currently we do not render videos/pictures that are attached to API Resources. 

<img width="1453" alt="Screen Shot 2022-07-29 at 12 52 40 PM" src="https://user-images.githubusercontent.com/35935196/181807523-47333e1e-b42c-4a79-b45e-52ba8fa9fe9a.png">

**Describe the solution you'd like**

If a file is representable on the browser (if its a video or a photo), render it.

PR #1030, opened 2022-08-18
[feature]: Render API Resource file attachment if representable (eg: video/ image)

alis-khadka:
Addresses: https://github.com/restarone/violet_rails/issues/980

https://user-images.githubusercontent.com/25191509/185351232-f54d0792-dcfc-4af7-ad05-ce7486da72da.mp4

### DIFF
--- a/app/helpers/api_resources_helper.rb
+++ b/app/helpers/api_resources_helper.rb
@@ -6,7 +6,7 @@ module ApiResourcesHelper
     output = ''
     output << <<-HTML
     #{image_tag (file.content_type.include?('image') ? file_url : ''), height: 200, width: 200, id: "#{label.parameterize.underscore}_preview_img", style: "display: #{file.content_type.include?('image') ? 'block' : 'none'};"}
-    #{video_tag (file.content_type.include?('image') ? file_url : ''), height: 200, controls: true, id: "#{label.parameterize.underscore}_preview_video", style: "display: #{file.content_type.include?('video') ? 'block' : 'none'};"}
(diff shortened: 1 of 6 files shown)


app/helpers/api_resources_helper.rb
```diff
@@ -1,12 +1,16 @@
...
 
-    if file.content_type.include?('image')
-      image_tag rails_blob_url(file, subdomain: Apartment::Tenant.current), height: 200, width: 200
-    else
-      link_to file.filename.to_s,  rails_blob_url(file, subdomain: Apartment::Tenant.current), target: '_blank'
-    end
+    file_url = rails_blob_url(file, subdomain: Apartment::Tenant.current)
+    output = ''
+    output << <<-HTML
+    #{image_tag (file.content_type.include?('image') ? file_url : ''), height: 200, width: 200, id: "#{label.parameterize.underscore}_preview_img", style: "display: #{file.content_type.include?('image') ? 'block' : 'none'};"}
```
donrestarone, line 8: @alis-khadka do we need to restrict the size of the preview? Can we show a responsive preview that makes better use of the space on the page? Please see below for an area marked in blue that is all space we can use for non-primitive property blocks and their previews

Screen Shot 2022-08-18 at 8 13 47 AM

alis-khadka: @donrestarone I have given max-height to the preview section. Let me know if its okay.

https://user-images.githubusercontent.com/25191509/185544841-a9590b58-c901-41ca-974e-45998209d5df.mp4

Issue #665, opened 2022-06-06
run sidekiq on Heroku free dyne


donrestarone:
https://bilalbudhani.com/running-sidekiq-on-heroku-free-dyno/

donrestarone:
https://www.reddit.com/r/ruby/comments/7fwxpa/running_sidekiq_on_heroku_free_dyno_with_puma/dqf0nwp/

PR #1032, opened 2022-08-19
[feature] Run sidekiq in forked process on Heroku dyno

alis-khadka:
Addresses: https://github.com/restarone/violet_rails/issues/665

## To run Sidekiq on Heroku free tier

set environment variable `INLINE_SIDEKIQ_WITH_PUMA` to `true` in Heroku.

### DIFF
--- a/Procfile
+++ b/Procfile
@@ -1,1 +1,2 @@
 release: bundle exec rails db:migrate
+web: bundle exec puma -C config/puma.rb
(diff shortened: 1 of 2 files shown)


donrestarone:
@alis-khadka how is this gated? On EC2, where we run sidekiq on a separate process-- would this spawn another sidekiq instance? 

That should not be the case. Sidekiq should run along the web process if ENV variable INLINE_SIDEKIQ_WITH_PUMA is set

alis-khadka:
> @alis-khadka how is this gated? On EC2, where we run sidekiq on a separate process-- would this spawn another sidekiq instance?
> 
> That should not be the case. Sidekiq should run along the web process if ENV variable INLINE_SIDEKIQ_WITH_PUMA is set

@donrestarone I have made the changes conditional to work with ENV variable INLINE_SIDEKIQ_WITH_PUMA.
Let me know if I am missing something.

PR #1070, opened 2022-08-28
Change cookies consent layout and add overlay

Ayon95:
Related to #1059 

Mobile:
cookies-consent-mobile

Desktop:
cookies-consent-desktop


app/views/cookies/index.html.erb
```diff
@@ -1,5 +1,6 @@
 <%= turbo_frame_tag :cookies_modal do %>
   <% if cookies[:cookies_accepted].nil? %>
-    <%= Subdomain.current.cookies_consent_ui.html_safe %>
+      <div class="cookies-consent__overlay"></div>
```
donrestarone, line 3: @Ayon95 adding this stub here makes customizing the consent popup more difficult than it should be. 

What if the client doesn't like the grayed out background? In this case it will be difficult to override it. 

Can you add the same functionality but only keep it to the migration? (as in remove this change)

db/migrate/20220824121504_add_cookies_consent_ui_to_subdomain.rb
```diff
@@ -1,12 +1,16 @@
 class AddCookiesConsentUiToSubdomain < ActiveRecord::Migration[6.1]
   COOKIE_CONSENT_UI = 
   "<div class=\"cookies-consent\">
-    <h2 class=\"cookies-consent__title\">We Value Your Privacy</h2>
-    <p>
-      We use cookies to enhance your browsing experience, serve personalized ads or content, and analyze our traffic. By clicking \"Accept All\", you consent to our use of cookies.
-    </p>
-    <a class=\"btn btn-primary mr-2 mb-2 mb-sm-0\" href=\"/cookies?cookies=true\">Accept All</a>
-    <a class=\"btn btn-outline-primary mb-2 mb-sm-0\" href=\"/cookies?cookies=false\">Reject All</a>
+    <div class=\"cookies-consent__text-content\">
```
donrestarone, line 4: @Ayon95 [per my comment here](https://github.com/restarone/violet_rails/pull/1070/files#r956800732)-- we should ideally only change the default migration value

app/assets/stylesheets/cookies.scss
```diff
@@ -4,22 +4,48 @@
 
 .cookies-consent {
   position: fixed;
-  bottom: 40px;
-  left: 40px;
-  width: 80%;
-  max-width: 550px;
+  bottom: 0;
+  left: 0;
```
donrestarone, line 8: @Ayon95 By adding the CSS here, wouldn't it be difficult for you to override from the client-side? Ideally we should have all the CSS for the default cookie consent in the migration. 

CC @Pralish for his thoughts

Ayon95: Should I specify a `style` attribute for the elements and move their styles there? The styles in media queries cannot be included in the `style` attribute.

db/migrate/20220824121504_add_cookies_consent_ui_to_subdomain.rb
Ayon95: Isn't this the default migration value?

app/assets/stylesheets/cookies.scss
donrestarone: @Ayon95 come to think of it, when you customize this you will use your own classes and markup right? In that case you can leave the styles here. 

just remove the `<div class="cookies-consent__overlay"></div>` and implement the background a different way without changing `app/views/cookies/index.html.erb`

db/migrate/20220824121504_add_cookies_consent_ui_to_subdomain.rb
donrestarone: @Ayon95 yup exactly. I was referring to not making the change here (`app/views/cookies/index.html.erb`)

Ayon95: Oh okay. I can remove the overlay from `app/views/cookies/index.html.erb` and include it in the default markup.

Ayon95:
Defined all styles in `style` attributes and moved the overlay from `app/views/cookies/index.html.erb` to the default markup.

Default markup:
```
<div class=\"cookies-consent__overlay position-fixed\" style=\"top: 0; bottom: 0; left: 0; right: 0; background-color: black; opacity: 0.5; z-index: 1000;\"></div>
  <div class=\"cookies-consent position-fixed bg-white d-md-flex justify-content-md-between\" style=\"bottom: 0; left: 0; width: 100%; padding: 2rem; z-index: 9000;\">
    <div class=\"cookies-consent__text-content col-md-8\" style=\"max-width: 700px;\">
      <h2 class=\"cookies-consent__title\" style=\"font-size: 1.4rem;\">We Value Your Privacy</h2>
      <p class=\"mb-4 mb-md-0\">
        We use cookies to enhance your browsing experience, serve personalized ads or content, and analyze our traffic. By clicking \"Accept All\", you consent to our use of cookies.
      </p>
    </div>
    <div class=\"cookies-consent__buttons-container d-flex flex-column col-md-4 col-xl-3\">
      <a class=\"btn btn-primary mb-3\" href=\"/cookies?cookies=true\">Accept All</a>
      <a class=\"btn btn-outline-primary\" href=\"/cookies?cookies=false\">Reject All</a>
    </div>  
  </div>
```

Mobile:
cookies-consent-mobile

Desktop:
cookies-consent-desktop

app/views/cookies/index.html.erb
```diff
@@ -1,5 +1,5 @@
 <%= turbo_frame_tag :cookies_modal do %>
   <% if cookies[:cookies_accepted].nil? %>
-    <%= Subdomain.current.cookies_consent_ui.html_safe %>
+      <%= Subdomain.current.cookies_consent_ui.html_safe %>
```
donrestarone, line 3: @Ayon95 you shouldn't have any changes on this file. You should do 

``` bash
git checkout Pralishcookies-consent -- app/views/cookies/index.html.erb

```

to make sure you reset the file back to how Pralish had it

Ayon95: Thanks. I've reset the file.

Issue #1073, opened 2022-08-30
GDPR/CCPA customer data request plugin


donrestarone:
This plugin will ingest user provides emails to send them the information they submitted information to the system. 

### config:
`name`: compliance/customer/request 
### properties:
`name`: Sally Anne
`slug`: compliance__customer_request 
`email`: email_address
`compliance_message_sent`: boolean

`exclude_api_namespaces`: array of `slug`'s of namespaces to exclude from lookup
`scan_all_namespaces`: boolean -- not required if `exclude_api_namespaces` is provided, but if `exclude_api_namespaces` is not provided the plugin will not work unless you acknowledge that lookup is allowed on all API namespaces by setting `scan_all_namespaces`: true
message: 'This is your data request, please see it below'

# Algorithm

When the plugin runs on a 1 minute cron schedule: 

1. find all `api_resources`  `compliance_message_sent`: false
2. for each `api_resource` look up all the namespaces allowed for look up based on `exclude_api_namespaces ` /  `scan_all_namespaces `-- list the api_resources that have the same property `email` and put them into its own CSV (by namespace)

## Parameters

Email message should be sent via the email system by creating a thread named: "#{email} compliance: customer submitted data request" and inserting the message there. 

The email message should contain the message defined in the message parameter along with a CSV for each of the API Namespaces that they have submitted data to (each API Namespace CSV will contain the resources that the user has submitted).

donrestarone:
# research

[See the seeds file](https://github.com/restarone/violet_rails/blob/master/db/seeds.rb) on how to programmatically send emails and things like that from the External API Connection.

[See here for an example](https://github.com/restarone/violet_rails/blob/master/test/plugins/mailchimp_plugin_test.rb) of a test for a Violet Rails plugin


library of plugins: https://github.com/restarone/violet_rails/blob/master/test/fixtures/external_api_clients.yml

PR #1082, opened 2022-09-03
Compliance visitor data request plugin

Ayon95:
Addresses #1073

### DIFF
--- a/app/mailers/e_mailer.rb
+++ b/app/mailers/e_mailer.rb
@@ -11,6 +11,15 @@ class EMailer < ApplicationMailer
     message_id = "#{Digest::SHA2.hexdigest(Time.now.to_i.to_s)}@#{Apartment::Tenant.current}@#{ENV['APP_HOST']}"
     @message_thread.update(current_email_message_id: message_id)
 
+    unless params[:attachments].nil?
+      Array.wrap(params[:attachments]).each do |attachment|
+        attachments[attachment[:filename]] = {
+          mime_type: attachment[:mime_type],
+          content: attachment[:content]
+        }
+      end
+    end
+
     mail(
       # This will make the mail addresses visible to all (no Blank Carbon Copy)
       to: @message_thread.recipients, 
--- a/app/mailers/e_mailer.rb
+++ b/app/mailers/e_mailer.rb
@@ -10,7 +10,8 @@ class EMailer < ApplicationMailer
             end
     message_id = "#{Digest::SHA2.hexdigest(Time.now.to_i.to_s)}@#{Apartment::Tenant.current}@#{ENV['APP_HOST']}"
     @message_thread.update(current_email_message_id: message_id)
-
+    
+    # if additional attachment params are passed, they are included in the email
     unless params[:attachments].nil?
       Array.wrap(params[:attachments]).each do |attachment|
         attachments[attachment[:filename]] = {
--- a/app/services/violet/connection_service.rb
+++ b/app/services/violet/connection_service.rb
@@ -1,0 +1,15 @@
+# Utitlity class for external API connection plugins
+class Violet::ConnectionService
+    def get_blob_url_with_current_subdomain(attachment)
+        Rails.application.routes.url_helpers.rails_blob_url(attachment, subdomain: Apartment::Tenant.current, host: ENV['APP_HOST'])
+    end
+
+    def get_current_subdomain_formatted_email
+        subdomain = Subdomain.current
+        if subdomain.email_name.present?
+            "#{subdomain.email_name} <#{subdomain.name}@#{ENV["APP_HOST"]}>"
+        else
+            "#{subdomain.name}@#{ENV["APP_HOST"]}"
+        end
+    end
+end
--- a/test/plugins/compliance_visitor_data_request_plugin_test.rb
+++ b/test/plugins/compliance_visitor_data_request_plugin_test.rb
@@ -17,7 +17,7 @@ class ComplianceVisitorDataRequestPluginTest < ActiveSupport::TestCase
         # By default, each visitor has no submission in an api namespace
     end
 
-    test "Should not send email to visitors if they already received one" do
+    test "Should not send email to visitor if they already received one" do
         @visitor_one.properties["compliance_message_sent"] = true
         @visitor_one.save
         @visitor_two.properties["compliance_message_sent"] = true
@@ -31,7 +31,7 @@ class ComplianceVisitorDataRequestPluginTest < ActiveSupport::TestCase
         end
     end
 
-    test "Should send email if visitor did not receive one and has made a submission" do
+    test "Should send email if compliance_message_sent is set to false on the API resource" do
         # Visitor one and two are making a submission under namespace_with_email_one
         ApiResource.create(api_namespace_id: @namespace_with_email_one.id, properties: {"post": "Hello world", "email": @visitor_one.properties["email"]})
         ApiResource.create(api_namespace_id: @namespace_with_email_one.id, properties: {"post": "Hello world", "email": @visitor_two.properties["email"]})
@@ -55,6 +55,8 @@ class ComplianceVisitorDataRequestPluginTest < ActiveSupport::TestCase
         end
 
         sent_email = ActionMailer::Base.deliveries.last
+        p Subdomain.current.name
+        p sent_email.from
         assert(sent_email.from.include?(Subdomain.current.name))
         assert(sent_email.from.include?(ENV["APP_HOST"]))
         assert_equal [visitor_email], sent_email.to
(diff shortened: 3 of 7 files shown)


test/plugins/compliance_visitor_data_request_plugin_test.rb
```diff
@@ -0,0 +1,159 @@
...
+
+        expected_message = @data_request_plugin.metadata["MESSAGE"]
+
+        perform_enqueued_jobs do
+            @data_request_plugin.run
+            Sidekiq::Worker.drain_all
+        end
+
+        sent_email = ActionMailer::Base.deliveries.last
+        assert(sent_email.from.include?(Subdomain.current.name))
```
Ayon95, line 58: It looks like this assertion is failing. For some reason, `Subdomain.current.name` is not being included. The test case passed when I ran it locally.

app/mailers/e_mailer.rb
```diff
@@ -11,6 +11,15 @@ def ship
     message_id = "#{Digest::SHA2.hexdigest(Time.now.to_i.to_s)}@#{Apartment::Tenant.current}@#{ENV['APP_HOST']}"
     @message_thread.update(current_email_message_id: message_id)
 
+    unless params[:attachments].nil?
```
donrestarone, line 14: @Ayon95 can you add a comment here, saying that:

`if additional attachment params are passed, they are included in the email`

app/services/violet/connection_service.rb
```diff
@@ -0,0 +1,15 @@
+# Utitlity class for external API connection plugins
+class Violet::ConnectionService
```
donrestarone, line 2: @Ayon95 can you rename this to `Violet::Connection`

app/services/violet/connection_service.rb
```diff
@@ -0,0 +1,15 @@
...
+    def get_blob_url_with_current_subdomain(attachment)
+        Rails.application.routes.url_helpers.rails_blob_url(attachment, subdomain: Apartment::Tenant.current, host: ENV['APP_HOST'])
+    end
+
+    def get_current_subdomain_formatted_email
```
donrestarone, line 7: rename to `get_subdomain_email_address`

app/services/violet/connection_service.rb
```diff
@@ -0,0 +1,15 @@
+# Utitlity class for external API connection plugins
+class Violet::ConnectionService
+    def get_blob_url_with_current_subdomain(attachment)
```
donrestarone, line 3: rename to `get_blob_url`

test/plugins/compliance_visitor_data_request_plugin_test.rb
donrestarone: @Ayon95 add a `p Subdomain.current.name` and  `p sent_email.from` statements on this line and read what's being printed in the CI log

test/plugins/compliance_visitor_data_request_plugin_test.rb
```diff
@@ -0,0 +1,159 @@
...
+        perform_enqueued_jobs do
+            @data_request_plugin.run
+            Sidekiq::Worker.drain_all
+        end
+
+        sent_email = ActionMailer::Base.deliveries.last
+        assert_equal expected_csv, sent_email.attachments.first.body.raw_source.gsub(/\r/, "")
+    end
+
+    test "Should raise an exception if visitor does not exclude any api namespace as well as does not give permission to scan all namespaces" do
```
donrestarone, line 146: `visitor` `user` `current_user` and `admin` are keywords. Lets not conflate them 

rename test name to: `Should raise an exception if admin does not exclude any api namespace and does not grant explicit permission to scan all API Namespaces`

test/plugins/compliance_visitor_data_request_plugin_test.rb
```diff
@@ -0,0 +1,159 @@
...
+        @namespace_with_email_two = api_namespaces(:namespace_with_email_two)
+
+        Sidekiq::Testing.fake!
+
+        # How test fixtures are set up:
+        # By default, "compliance_message_sent" of each visitor api resource is false
+        # By default, each visitor has no submission in an api namespace
+    end
+
+    test "Should not send email to visitors if they already received one" do
```
donrestarone, line 20: `Should not send email to visitor if they already received one` -- `visitor` -- **singular**

test/plugins/compliance_visitor_data_request_plugin_test.rb
```diff
@@ -0,0 +1,159 @@
...
+
+        assert_no_difference "ActionMailer::Base.deliveries.size" do
+            perform_enqueued_jobs do
+                @data_request_plugin.run
+                Sidekiq::Worker.drain_all
+            end
+        end
+    end
+
+    test "Should send email if visitor did not receive one and has made a submission" do
```
donrestarone, line 34: rename test to: `Should send email if compliance_message_sent is set to false on the API resource`

test/plugins/compliance_visitor_data_request_plugin_test.rb
```diff
@@ -0,0 +1,159 @@
...
+        # Should not send an email even if a submission was made under a namespace since it's excluded
+        assert_no_difference "ActionMailer::Base.deliveries.size" do
+            perform_enqueued_jobs do
+                @data_request_plugin.run
+                Sidekiq::Worker.drain_all
+            end
+        end
+    end
+
+    test "Email should contain the right number of CSV attachments" do
```
donrestarone, line 78: rename to: `Email should contain a CSV attachment per API Namespace`

test/plugins/compliance_visitor_data_request_plugin_test.rb
```diff
@@ -0,0 +1,159 @@
...
+            @data_request_plugin.run
+            Sidekiq::Worker.drain_all
+        end
+
+        # Visitor one making submissions under two namespaces, so the email should contain two attachments
+        sent_email = ActionMailer::Base.deliveries.last
+        assert_equal 2, sent_email.attachments.length
+
+        sent_email.attachments.each do |attachment|
+            assert(attachment.content_type.start_with?("text/csv"))
```
donrestarone, line 93: can you also assert that the two CSV's are different (since they are pertaining to a different API Namespace) and include the data for its respective API Namespace (eg: one API namespaces has attributes: `name`,` email`, `phone` -- while the other has `address` and `patient_id`

test/plugins/compliance_visitor_data_request_plugin_test.rb
```diff
@@ -0,0 +1,159 @@
...
+        end
+    end
+
+    test "CSV file should have the right format" do
```
donrestarone, line 97: you are not testing format in this test case, so it should be renamed to "CSV file should be named according to the API Namespace and have file extension .csv"

test/plugins/compliance_visitor_data_request_plugin_test.rb
Ayon95: Like this?
```ruby
assert_not(first_attachment_content == second_attachment_content)

@namespace_with_email_one.properties.each do |key, value|
   assert(first_attachment_content.include?(key))
   assert(first_attachment_content.include?(value))
end

@namespace_with_email_two.properties.each do |key, value|
  assert(second_attachment_content.include?(key))
  assert(second_attachment_content.include?(value))
end
```

test/plugins/compliance_visitor_data_request_plugin_test.rb
Ayon95: `sent_email.from` is returning an array containing a single email address.

https://github.com/restarone/violet_rails/runs/8239212970?check_suite_focus=true#step:14:70

```
...................................................................................."public"
["public@restarone.com"]
F

Failure:
ComplianceVisitorDataRequestPluginTest#test_Sender_and_recipient_email_addresses_and_body_message_are_correct [/home/runner/work/violet_rails/violet_rails/test/plugins/compliance_visitor_data_request_plugin_test.rb:60]:
Expected false to be truthy.
```

test/plugins/compliance_visitor_data_request_plugin_test.rb
donrestarone: @Ayon95 yup like that

test/plugins/compliance_visitor_data_request_plugin_test.rb
donrestarone: @Ayon95 were you able to resolve this? Im not sure why its returning 2 different values, but a hack could be to array wrap it 

 `Array.wrap(sent_email.from).include?(Subdomain.current.name)`

Ayon95: Yeah I was also thinking of using `Array.wrap`. Wanted to hear your thoughts on how to handle this.

donrestarone:
@Ayon95 is this ready for showtime?

Ayon95:
Emails get sent with the CSV attachments properly. The message thread that gets created in Email in admin does not have the attachments. Will this be an issue?
Capture

Ayon95:
Message thread with attachments:
Capture2
Capture

Issue #1052, opened 2022-08-24
Dynamic links get broken in send_email API action once changes are saved


Ayon95:
**Describe the bug**
When a link with dynamic segments is added to send_email API action, its URL gets broken after hitting Save. To prevent the link from getting broken, we have to write its markup every time we update API actions.

**To Reproduce**
Add a send_email API action and insert a link with a dynamic part
![Capture](https://user-images.githubusercontent.com/73725297/186510787-e636f20b-3ee2-4e6d-aeb8-aa4388b1a340.PNG)

Make some changes and hit Save:
![Capture2](https://user-images.githubusercontent.com/73725297/186511439-cf4af744-b3e4-4166-95b5-12370c10b2f9.PNG)

**Expected behavior**
The dynamic URL should remain unchanged and still work after making changes.

**Additional context**
The URL `https://google.com?email=<%= api_resource.properties['email'] %>` gets changed to `https://google.com/?email=%3C%=%20api_resource.properties%5B%27email%27%5D%20%%3E`

The following error is thrown when the corresponding API action ID is clicked:
![screencapture-lvh-me-5250-api-namespaces-6-api-actions-60-2022-08-24-15_49_53](https://user-images.githubusercontent.com/73725297/186513116-f4716d92-90a8-49cd-9f16-7a1be0bfa918.png)

```
(erb):3: unknown type of %string
; _erbout.<<((%20api_resource.properties%5B'em...
^~~
(erb):3: syntax error, unexpected end-of-input
; _erbout.<<((%20api_resource.properties%5B'em...
^~~
```

donrestarone:
related: https://github.com/restarone/violet_rails/issues/1045

donrestarone:
<img width="1117" alt="Screen Shot 2022-08-29 at 8 37 34 AM" src="https://user-images.githubusercontent.com/35935196/187202868-3125fd94-1ba4-412e-9ab7-a9d0350ff904.png">

PR #1085, opened 2022-09-05
decode uri-encoded characters

Pralish:
Addresses: https://github.com/restarone/violet_rails/issues/1052

Demo:

https://user-images.githubusercontent.com/50227291/188492173-82f9a8ed-609a-44b1-a608-77aa0e1d1600.mov


donrestarone:
@Pralish this fixes the server side, but when Trix breaks the code the encoded characters are still shown to the user. 

Can we prevent this case?

donrestarone:
@Pralish to handoff to a client engineer to explore fixing trix via JS (if there are ERB tags don't URL encode the string)

Issue #1071, opened 2022-08-29
sidekiq dead set functionality is broken


donrestarone:
<img width="864" alt="Screen Shot 2022-08-29 at 9 54 41 AM" src="https://user-images.githubusercontent.com/35935196/187217741-78e02f11-ac91-44d5-a690-e1d7d8c561c6.png">

the yield here: [app/models/api_resource.rb:80](https://github.com/restarone/violet_rails/blob/master/app/models/api_resource.rb#L80) is causing sidekiq jobs to get stuck in a `failed` state -- this should not be the case. We should instead see the jobs show up under `dead`. `Failed` is simply a transitive state. 

https://github.com/mperham/sidekiq/issues/2760#issuecomment-170112610

Pralish:
@donrestarone , Can you confirm if the stuck jobs are actually api_actions stuck in 'executing' state?

donrestarone:
@Pralish yup! They are stuck in "executing" state

Issue #985, opened 2022-07-30
run error actions (if any) when recaptcha fails


donrestarone:
**Is your feature request related to a problem? Please describe.**

Currently we don't fire error actions when recaptcha fails

**Describe the solution you'd like**

fire error actions when recaptcha fails

<img width="1728" alt="Screen Shot 2022-07-30 at 1 21 42 PM" src="https://user-images.githubusercontent.com/35935196/181934704-db882ca1-0684-47cb-9394-790ca846e875.png">

donrestarone:
@Pralish while you're looking into this, can you address this issue? Its related to how we are handling errors. https://github.com/restarone/violet_rails/issues/1071

PR #1086, opened 2022-09-06
Run Error api actions

Pralish:
Addresses: https://github.com/restarone/violet_rails/issues/985

Addresses: https://github.com/restarone/violet_rails/issues/1071

### Demo:

https://user-images.githubusercontent.com/50227291/188673867-186d03e9-a8ec-4cb7-adb0-6a6483ba1cec.mov

Since api resource is not saved due to any error, api action will keep track of user input with proper error message under 

`additional_data['api_resource']['properties']` and `additional_data['api_resource']['errors']`

Demo shows error message and user input being accessed by api_action

### Demo fix dead set functionality

https://user-images.githubusercontent.com/50227291/189976658-e7c6fbe8-1f69-4740-9e98-98835a8fae26.mov

### DIFF
--- a/app/models/concerns/dynamic_attribute.rb
+++ b/app/models/concerns/dynamic_attribute.rb
@@ -22,8 +22,7 @@ module DynamicAttribute
         end
 
         # parse erb
-        # unescape html and decode uri-encoded charcters
-        parser = ERB.new(CGI.unescapeHTML(URI.decode(parsed_text)))
+        parser = ERB.new(CGI.unescapeHTML(parsed_text))
         result = parser.result(binding)
         result.html_safe
       end
--- a/app/models/concerns/dynamic_attribute.rb
+++ b/app/models/concerns/dynamic_attribute.rb
@@ -22,7 +22,8 @@ module DynamicAttribute
         end
 
         # parse erb
-        parser = ERB.new(CGI.unescapeHTML(parsed_text))
+        # unescape html and decode uri-encoded charcters
+        parser = ERB.new(CGI.unescapeHTML(URI.decode(parsed_text)))
         result = parser.result(binding)
         result.html_safe
       end
--- a/app/models/concerns/dynamic_attribute.rb
+++ b/app/models/concerns/dynamic_attribute.rb
@@ -9,7 +9,8 @@ module DynamicAttribute
     include Rails.application.routes.url_helpers
 
     included do
-      def parse_dynamic_attribute(value)
+      def parse_dynamic_attribute(value, context = {})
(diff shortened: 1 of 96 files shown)


app/models/concerns/dynamic_attribute.rb
```diff
@@ -22,7 +22,8 @@ def parse_dynamic_attribute(value)
         end
 
         # parse erb
-        parser = ERB.new(CGI.unescapeHTML(parsed_text))
+        # unescape html and decode uri-encoded charcters
```
donrestarone, line 25: @Pralish this change is not supposed to be here (referenced: https://github.com/restarone/violet_rails/pull/1085)

donrestarone:
@Pralish does this PR also address https://github.com/restarone/violet_rails/issues/1071 ? It was mentioned in the original task since you'd touch code related to it

app/models/concerns/dynamic_attribute.rb
Pralish: @donrestarone Thanks for noticing. I must have checked out from the referenced branch instead of master.

Pralish:
> @Pralish does this PR also address #1071 ? It was mentioned in the original task since you'd touch code related to it

@donrestarone , This PR doesn't. I have looked into the issue and was planning to work on a separate PR.

donrestarone:
@Pralish sounds good. I'll assign the task to you to pick up later

db/migrate/20220906063031_add_additional_data_to_api_actions.rb
```diff
@@ -0,0 +1,5 @@
+class AddAdditionalDataToApiActions < ActiveRecord::Migration[6.1]
+  def change
+    add_column :api_actions, :additional_data, :jsonb, default: {}
```
donrestarone, line 3: rename to `meta_data`

donrestarone:
@Pralish can you address the feedback here? https://github.com/restarone/violet_rails/pull/1086#discussion_r972509593 -- after that we can deploy it for testing!

Pralish:
@donrestarone , renamed the attribute.

I also found an issue. We run error actions incase of any errors. Since we run api actions in background, incase of errors, Sidekiq reruns the job. Error action gets triggered each time the job fails.  We have retries count of 3, so if the job fails all three times , error actions get executed 3 time. 

Shouldn't error action run only after final retry?

donrestarone:
@Pralish if it's a small change yeah let's do it. If not, let's leave as is-- since we are currently not even able to re-run failing jobs

PR #1093, opened 2022-09-11
Search forum complete

peaceofmind123:
Implemented search functionality in /forum and associated test case. A significant problem I ran across was validating whether the full text search field :body is empty or not. After having added "has_one :body, class_name: 'ActionText::RichText', as: :record" on forum_post model, the validation of disallowing empty body string passed even when the body was an empty string. Since  ActionText::Richtext has no in-built functionality for empty full text validation, (as mentioned in https://stackoverflow.com/questions/64580428/is-it-possible-to-prevent-empty-action-text-entries ), I used the solution the question offered: handling the validation in controller create. During the creation of the forum_post, I check if it is empty, and if it is, omit the body altogether. Although not ideal, it works as expected and all tests are passing, as is evident from the second video below.

https://user-images.githubusercontent.com/13151686/189536951-44f6c421-2349-404c-af22-81d1b7753eef.mp4

https://user-images.githubusercontent.com/13151686/189537366-ed156007-151f-4f8f-9508-0896352c6e83.mp4

### DIFF
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -552,7 +552,7 @@ DEPENDENCIES
   will_paginate
 
 RUBY VERSION
-   ruby 2.6.6p146
+   ruby 3.0.0P0
 
 BUNDLED WITH
    2.1.4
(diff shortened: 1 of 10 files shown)


donrestarone:
@peaceofmind123 thanks for your coding challenge submission! I have created a base branch for you on this repo so your change can be deployed to our testnet

PR #1100, opened 2022-09-12
Peaceofmind123search forum complete

donrestarone:


donrestarone:
clicking on category on a thread results in error, indicating that we have a gap in our test suite (already tracked by: https://github.com/restarone/violet_rails/issues/111)
Screen Shot 2022-09-13 at 7 23 09 PM
Screen Shot 2022-09-13 at 7 23 00 PM

Issue #1110, opened 2022-09-15
cannot use the same API form (recaptcha v3) multiple times on the same page


donrestarone:
We have a form (see marked in blue) that is used multiple times on the same page. Only the first form works, the rest of them are broken

![Screenshot 2022-09-14 at 21-44-50 ordinarytrip com](https://user-images.githubusercontent.com/35935196/190294203-283c0604-1365-440c-80a0-6c37cda370b0.png)

PR #1114, opened 2022-09-16
[bug fix] resolve use API form (recaptcha v3) multiple times on the same page

donrestarone:
[bug fix] resolve use API form (recaptcha v3) multiple times on the same page

Addresses: https://github.com/restarone/violet_rails/issues/1110

### DIFF
--- a/app/controllers/concerns/api_actionable.rb
+++ b/app/controllers/concerns/api_actionable.rb
@@ -152,6 +152,7 @@ module ApiActionable
 
   def render_fallback_to_recaptcha_v2_with_error_message(error_message)
     @flash = { error: error_message }
+    @form_id = params[:form_id]
     render 'shared/fallback_to_recaptcha_v2.js.erb'
   end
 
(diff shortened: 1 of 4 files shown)


donrestarone:
@Pralish This addresses the issue. But re-render recaptcha doesn't work Can you fix that regression? (I waited for the recaptcha challenge to expire and was not able to get a fresh challenge)


⚠️ temporary deployment underway to ordinarytrip.com -- heads up @westernal you are now unblocked on the forms.

Pralish:
@donrestarone , I didn't quite understand the issue. But I was able to replicate this one. Is this what you are talking about.


https://user-images.githubusercontent.com/50227291/190912375-f3c5e53f-0f1a-43ab-a985-f759beb763a6.mov


I waited for the challenge to expire, submitted the form, it rerendered v2, I completed the challenge, but the button didn't get enabled

donrestarone:
@Pralish thats similar to the case I was referring to. 

The scenario is when the solved challenge expires, eg: the user completes the challenge and waits for a while. When it expires the new challenge isn't re-rendered -- the form is simply locked up

Issue #1084, opened 2022-09-05
Unique string aggregator plugin 


donrestarone:
given a [JSON path](https://support.smartbear.com/alertsite/docs/monitors/api/endpoint/jsonpath.html) that points to a string value, this plugin will scan the the associated API Resources of the API Namespace it is run against to output a unique list of API Resources to a target API Namespace

# Namespace

NA since it runs on an existing Namespace

# Properties

`resolve_string_path`: "api_namespace_slug.some_property.another_property"
`output_api_namespace`: "api_namespace_slug"
`pristine`: boolean 

# Algorithm

## Validations

1. `resolve_string_path`'s `api_namespace_slug` should be the same API Namespace the External API Connection is running on. 
2. raise exception "API Namespace resource pollution detected, check resolve_string_path and output_api_namespace" if `resolve_string_path`'s "api_namespace_slug" and `output_api_namespace`'s "api_namespace_slug" are the same
3. if `pristine` is set to` true`, collect all ID's of all existing API Resources in the target API Namespace and delete them *after* successful connection execution

## Logic

1.  if `pristine` is set to` true`:  calculate all the ID's of the API resources to delete in the target API Namespace, assign that to a variable and remove them after the algorithm runs 

2. Access the `resolve_string_path` in the current API Namespace, and loop over all the API Resources building a hash. 
3. Take the hash and loop over the keys creating an API Resource in the target namespace for each key.  The created API Resources should have the following format:  
`property_name` (derived form "api_namespace_slug.some_property.another_property" in`resolve_string_path` )

``` javascript
{
`property_name`: "string value"  
}
```
in the example `property_name` is the resulting property of the JSON path evaluation in `"api_namespace_slug.some_property.another_property"`

## Example

API Namespace: Movies

``` javascript
title: 'titanic',
tags: ['romance', 'shipbuilding']
```

given an API Namespace of movies, build a unique list of tags that each movie is tagged under.

donrestarone:
@Sushmit1997 have you had a chance to take a look at this?

Sushmit1997:
@donrestarone Yes, I'm looking at this. Will post estimates shortly.

Sushmit1997:
> @Sushmit1997 have you had a chance to take a look at this?

I've posted estimate for this task in my backlog.

Sushmit1997:
@donrestarone 

In logic number 1, what does `assign that to a variable` means? 
And I didn't get logic number 3

CC @Pralish @alis-khadka

[...]

PR #1133, opened 2022-09-26
Unique string aggregator plugin

Ayon95:
Addresses #1084 

https://user-images.githubusercontent.com/73725297/192399784-599d236c-86b3-41f0-8f2c-8f27781f6b72.mp4

Pristine mode:

https://user-images.githubusercontent.com/73725297/192399810-05c59865-2f54-400c-afa3-47868b5fe439.mp4

### DIFF
--- a/test/fixtures/api_namespaces.yml
+++ b/test/fixtures/api_namespaces.yml
@@ -155,3 +155,25 @@ no_api_actions:
               }
   requires_authentication: false
   namespace_type: create-read-update-delete
+
+movies:
+  name: movies
+  slug: movies
+  version: 1
+  properties: {
+                "title": "",
+                "tags": []
+              }
+  requires_authentication: false
+  namespace_type: create-read-update-delete
+
+movie_tags:
+  name: movie_tags
+  slug: movie_tags
+  version: 1
+  properties: {
+                "tags": "",
+                "representation": ""
+              }
+  requires_authentication: false
+  namespace_type: create-read-update-delete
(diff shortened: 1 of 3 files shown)


donrestarone:
@Ayon95 has this been tested on `nikean.org`?

Ayon95:
I haven't tested it on nikean.org yet. I had a question regarding `PRISTINE` metadata property. What should the plugin do if `PRISTINE` is set to false?

donrestarone:
@Ayon95 if `pristine` is set to false it shouldn't remove the existing data set. 

if `pristine` is set to true it will replace the API resources in a ZDT (zero down time) way (as described in the task)

Ayon95:
Okay, thanks for clarifying. That's what the plugin does currently.

donrestarone:
@Ayon95 I just merged your PR for the compliance plugin, can you fix the merge conflicts on this branch?

Ayon95:
Added documentation.

donrestarone:
@Ayon95 Have we considered the following flow: 

1. an admin adds a new story 
2. runs the connections for generating the tags

Do they have to manually fix the representations because they added 1 more piece of content?

If so we should change the algorithm such that its additive when `pristine` is not selected. It should create any tags that are missing (with no duplication). If `pristine` is set, then it should delete everything and rebuild from scratch. 

This way, we need to only adjust the representation for the tag that was newly created

Ayon95:
Okay, I'll make the change.

donrestarone:
@Ayon95 the the changes in place we should update the connection code on ‘Nikean.org‘ before we make any changes to the video library

Ayon95:
Agreed. Should `PRISTINE` be set to false?

donrestarone:
Yeah let's leave pristine as false @Ayon95

Issue #1137, opened 2022-09-28
Ability to query API resources that satisfy any one of multiple conditions


Ayon95:
**Is your feature request related to a problem? Please describe.**
Currently, it's only possible to specify queries with multiple conditions where all conditions need to be met.

https://github.com/restarone/violet_rails/blob/master/test/models/concerns/jsonb_search/query_builder_test.rb#L29

**Describe the solution you'd like**
There should be a way to query API resources by multiple conditions where only one condition needs to be true.

**Additional context**
This feature can be used to implement broad-scoped searching (look for a search term in multiple properties) with the [API HTML renderer](https://github.com/restarone/violet_rails/pull/1108)

PR #1145, opened 2022-10-04
[feature] API resource query match any one condition

Regaron:
Addresses: [#1137](https://github.com/restarone/violet_rails/issues/1137)

Screenshot:
Screenshot 2022-10-04 at 4 37 24 PM
Screenshot 2022-10-04 at 4 37 03 PM

Passing the third argument, ``'ANY'``  in ``build_jsonb_query`` or ``jsonb_search`` method will generate the clause requiring only one condition to be satisfied.

### DIFF
--- a/app/helpers/content_helper.rb
+++ b/app/helpers/content_helper.rb
@@ -27,9 +27,9 @@ module ContentHelper
 
     response = response.where.not(user_id: nil).where(user_id: current_user&.id) if scope&.dig('current_user') == 'true'
 
-    response = response.jsonb_search(:properties, scope["properties"]) if scope["properties"]
+    response = response.jsonb_search(:properties, scope["properties"], scope["match"]) if scope["properties"]
 
-    response = response.jsonb_search(:properties, JSON.parse(params[:properties]).to_hash) if params[:properties]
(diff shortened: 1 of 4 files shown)


donrestarone:
@Regaron i see a bunch of changes that are no-op's-- can you remove this noise and isolate the change to the code you actually changed?

Regaron:
@donrestarone, I've removed the no-ops.

donrestarone:
Looking good @Regaron  -- can you add documentation for client engineers to use this functionality? [See here for docs on the renderer](https://github.com/restarone/violet_rails/releases/tag/0.9.43) -- we will need an example like that here

CC @Ayon95

Regaron:
### Demo

https://user-images.githubusercontent.com/18470532/194049679-a9ea69cf-4af7-4902-a371-31a51c7ae0bf.mov

https://user-images.githubusercontent.com/18470532/194049874-3987e65b-b656-4b51-91f1-13e3476018ef.mp4

donrestarone:
@Ayon95 yielding to you to verify that this is what you need for `Nikean.org`

A good test would be to add searching by multiple attributes to the car app we built at http://violetrails.herokuapp.com

Ayon95:
✔ UAT passed: query API resources that match any one of multiple conditions

PR #1150, opened 2022-10-09
[feature] Unique string aggregator plugin (#1133) (#1143)

donrestarone:
Addresses #1084

This plugin takes an array of strings and creates a list of API resources (one per unique string value) for an output API namespace. For example, for two arrays - `tags: ["animation", "family", "martial arts"]`,`tags: ["animation", "family"]`  - in `movies` API namespace, 3 API resources will be created under `movie_tags` (the output API namespace in this case):

```ruby
{ "tags" => "animation", "representation" => "Animation" },
{ "tags" => "family", "representation" => "Family" },
{ "tags" => "martial arts", "representation" => "Martial Arts" }
```

https://user-images.githubusercontent.com/73725297/192908391-21560b30-1977-4451-ae15-5564d4289eb4.mp4

## Metadata

The following metadata properties are required:

- `INPUT_PROPERTY`: the name of the property that is mapped to an array of strings in the input API namespace (the API namespace that the plugin will run against)
- `OUTPUT_API_NAMESPACE`: the slug of the API namespace under which API resources will be created
- `PRISTINE`: a boolean indicating whether or not the plugin will run in pristine mode

If `PRISTINE` is true, then existing API resources in the output API namespace will be deleted after creating new API resources. If `PRISTINE` is false, then old API resources will not be deleted and new ones will be created only for new unique strings.

metadata

### Pristine mode

https://user-images.githubusercontent.com/73725297/192908436-351d44b3-c738-4973-b851-7f2529008aed.mp4

### Non-pristine mode

https://user-images.githubusercontent.com/73725297/193096865-63780536-dad0-4753-82ef-1e834c1ed362.mp4

## Shape of the output API namespace

Each API resource of the output API namespace must have the following two properties:

- `<INPUT_PROPERTY>`: this property must be the same as `INPUT PROPERTY` provided in the plugin metadata
- `representation`: the value of this property can be used as the display text for an `<option></option>` HTML element

output-api-namespace

### DIFF
--- a/test/fixtures/api_namespaces.yml
+++ b/test/fixtures/api_namespaces.yml
@@ -189,3 +189,25 @@ namespace_with_email_two:
   }
   requires_authentication: false
   namespace_type: create-read-update-delete
+  
+movies:
+  name: movies
+  slug: movies
+  version: 1
+  properties: {
+                "title": "",
+                "tags": []
+              }
+  requires_authentication: false
+  namespace_type: create-read-update-delete
+
+movie_tags:
+  name: movie_tags
+  slug: movie_tags
+  version: 1
+  properties: {
+                "tags": "",
+                "representation": ""
+              }
+  requires_authentication: false
+  namespace_type: create-read-update-delete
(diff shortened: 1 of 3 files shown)


donrestarone:
⚠️  plugin + test change, no deploy required

Issue #1149, opened 2022-10-09
videos not playing on safari


donrestarone:
**Describe the bug**
videos do not play on safari

**To Reproduce**
1. Create an API Namespace with a file upload 
2. upload a video
3. use renderer to render the video with an HTML video element
4. try playing the video on safari5. 
5. video does not play
**Expected behavior**
video plays

the following error is in the console: 
``` txt
The page at https://nikean.org/story-details?id=961392 was allowed to display insecure content from http://public.nikean.org/rails/active_storage/blobs/redirect/eyJfcmFpbHMiOnsibWVzc2FnZSI6IkJBaHBBa0VCIiwiZXhwIjpudWxsLCJwdXIiOiJibG9iX2lkIn19--d97d3e3a946684c589fe4b6bcf6133b4518e94bc/Sanjay%20-%20Long%20Form.m4v.

```

**Screenshots**

<img width="1728" alt="Screen Shot 2022-10-08 at 8 12 23 PM" src="https://user-images.githubusercontent.com/35935196/194731890-c5c8c413-2dfb-4168-930e-1bec6a7688db.png">

**Desktop (please complete the following information):**
 - OS: MacOS
 - Browser Safari

donrestarone:
a possible fix would be to load assets over `https` since the error is complaining that assets are being served over `http`

PR #1156, opened 2022-10-13
Load assests over https

Regaron:
Addresses: [#1149](https://github.com/restarone/violet_rails/issues/1149)
- Add config to load assets over https for staging and production.


donrestarone:
@Regaron elegant change, what about the blob URL's in https://github.com/restarone/violet_rails/blob/feaf01fffbf13fcff03f004544ff297a6d155639/app/views/active_storage/blobs/_blob.html.erb ?

Regaron:
@donrestarone I changed it to be based on config to work everywhere url_helpers are used.

PR #1168, opened 2022-10-20
[DO NOT MERGE] Remove api clients

Pralish:


donrestarone:
we should prioritize this after entity analytics: https://github.com/restarone/violet_rails/pull/1332

Issue #1127, opened 2022-09-24
migrate testnet off Heroku


donrestarone:
Currently we test every change on http://violetrails.herokuapp.com before releasing it to our customers. We have enjoyed Heroku's generous free tier for testing in the last year. 

Starting November 28, 2022 Heroku will shut down free services. In preparation, we need to migrate the testnet off Heroku and into a PaaS hosted on AWS EC2

1. Export the DB: https://devcenter.heroku.com/articles/heroku-postgres-import-export#export
2. Setup PaaS using Dokku: https://github.com/dokku/dokku 
3. change Heroku deploy script to deploy to Dokku: https://github.com/restarone/violet_rails/blob/master/.github/workflows/heroku-deploy.yml 
4. create a Dokku app for each open PR on the Dokku server (using subdomains) -- each push to that PR should deploy to the respective Dokku app on the Dokku server

## Dokku

running Rails on Dokku: 

1. https://www.pedroalonso.net/blog/hosting-rails-in-dokku
2. https://majesticlabs.dev/the-ultimate-guide-to-dokku-and-ruby-on-rails-5/

donrestarone:
## AWS Setup

EC2 instance with elastic IP connected to route 53 with port 22 open. 


## install Dokku
run this on the EC2 instance to install Dokku
``` bash
# for debian systems, installs Dokku via apt-get
wget https://raw.githubusercontent.com/dokku/dokku/v0.28.1/bootstrap.sh
sudo DOKKU_TAG=v0.28.1 bash bootstrap.sh
```

add your admin user, run this on your local machine-- it will copy your public key to the Dokku EC2 server so you can push your application to it

``` bash
cat ~/.ssh/id_rsa.pub | ssh SERVER_IP_HERE sudo dokku ssh-keys:add admin
```

donrestarone:
add `taron.cloud` to the automated deployment pipeline

PR #1183, opened 2022-10-28
[Dokku CI] create separate redis service for the review-apps in dokku-CI

alis-khadka:
Address: https://github.com/restarone/violet_rails/issues/1127


alis-khadka:
@donrestarone Since, the dokku-CI won't run in, could you check this PR by creating followup PR on the main repo?

Issue #974, opened 2022-07-26
link to associated CMS entity in API


donrestarone:
**Is your feature request related to a problem? Please describe.**
currently there is no way to tell if an API Namespace is being used in the CMS (for form rendering)

**Describe the solution you'd like**

in a column in the index view show if there are associations (has CMS associations :heavy_check_mark: ) and in the API Namespace `#show` view display the associated CMS Pages/Snippets (each association should be clickable leading to the the respective associated entity's `#show` page)
<img width="1728" alt="Screen Shot 2022-07-26 at 6 57 39 PM" src="https://user-images.githubusercontent.com/35935196/181126508-f6f35b0f-ca14-44c9-8edc-22e37767d6c8.png">

we need to detect the following token in Snippets, Pages and Layouts
<img width="456" alt="Screen Shot 2022-07-26 at 7 00 22 PM" src="https://user-images.githubusercontent.com/35935196/181126655-b312f1a4-847c-4564-b321-cecaea20bdeb.png">
example usage:
<img width="1208" alt="Screen Shot 2022-07-26 at 7 00 47 PM" src="https://user-images.githubusercontent.com/35935196/181126658-36fcca00-8a8f-4100-88ed-2bcb26ac0f3d.png">
in the case above the API Namespace should link back to the snippet `show` view

it should also detect HTML rendering:

``` ruby
{{ cms:helper render_api_namespace_resource_index 'cars', scope: { properties:  { published: 'true' }, order: { created_at: 'DESC' } } }}
```

PR #1189, opened 2022-11-01
[feature] link to associated CMS entity in API

donrestarone:
Addresses: https://github.com/restarone/violet_rails/issues/974

**Demo**

https://user-images.githubusercontent.com/25191509/199165299-25d1e0fb-0cd1-49a4-8e66-224bca2fa660.mov

### DIFF
--- a/app/controllers/comfy/admin/api_namespaces_controller.rb
+++ b/app/controllers/comfy/admin/api_namespaces_controller.rb
@@ -1,3 +1,5 @@
+require 'will_paginate/array'
+
 class Comfy::Admin::ApiNamespacesController < Comfy::Admin::Cms::BaseController
   before_action :ensure_authority_to_manage_api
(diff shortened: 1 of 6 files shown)


donrestarone:
@alis-khadka please take a look at the review app, the association for the renderer is not showing up: 
Screen Shot 2022-11-01 at 5 52 02 PM
Screen Shot 2022-11-01 at 5 51 56 PM

We expect to see a snippet association because we are using the snippet to render the API Namespace (note that there is a snippet for the show page and the index page per: https://github.com/restarone/violet_rails/releases/tag/0.9.43)

alis-khadka:
Here is the updated demo video for this:


https://user-images.githubusercontent.com/25191509/199496018-dfda3568-4de6-469e-9ecc-95c0a8d082a3.mov

donrestarone:
@alis-khadka its still not showing certain pages, like this one:

Screen Shot 2022-11-02 at 5 46 53 PM

Can you check the review app? See under the car app

Issue #1144, opened 2022-10-01
prevent raw SQL usage in API 


donrestarone:
**Is your feature request related to a problem? Please describe.**

currently It is possible to run raw SQL queries in an External API Connection or a Custom Action. 

https://stackoverflow.com/questions/14824453/rails-raw-sql-example

**Describe the solution you'd like**

in app/models/concerns/safe_executable_validator.rb prevent usage of: 

1. `Activerecord::Base`
2. `*.find_by_sql`
3. `Rails`

donrestarone:
to be clear, raw SQL means complete SQL statements not fragments like this: https://guides.rubyonrails.org/active_record_querying.html#array-conditions

PR #1220, opened 2022-11-11
[feature] Prevent raw sql usage in API

donrestarone:
Addresses: https://github.com/restarone/violet_rails/issues/1144


donrestarone:
note: after deployment `UniqueStringAggregator` and `VacuumJob` both need to be updated with the code in test/fixtures/external_api_clients.yml

PR #1167, opened 2022-10-19
Web hook drive strategy + Decouple api clients

Pralish:
Addresses: https://github.com/restarone/violet_rails/issues/682

Api Keys:

https://user-images.githubusercontent.com/50227291/196992417-ab91ed2e-1514-4abf-9954-d44123542583.mov

Webhook:

https://user-images.githubusercontent.com/50227291/196992487-4c3c9593-4cb6-4b8c-8f7b-c6e22d9385a2.mov

**The webhook's payload can be accessed as `parameters[:request]['body']` in external api client.**

Eg: https://github.com/restarone/violet_rails/pull/1167/files#diff-1aa9cadc4b4c76bf7ec3d7c0cf8d0da66be3ba87336bad4d32241c99d53aa6beR603-R623

```
    class ExternalApiModelExample
      def initialize(parameters)
        @external_api_client = parameters[:external_api_client]
        @request = parameters[:request]
      end
  
      def start
        if @request['body']['type'] == 'customer.created'
          @external_api_client.api_namespace.api_resources.create(
            properties: {
              request_body: @request["body"]	
            }
          )
        end
      end
```

Custom webhook verification method:

https://user-images.githubusercontent.com/50227291/197921958-a06b85a8-7c7a-4e3b-afb9-889fe5c6e110.mov

Issue #682, opened 2022-06-07
ability to receive web hooks (both authenticated and unauthenticated) and map them to External API Connections


donrestarone:
## Add drive strategy: web-hook to External API Connection
<img width="1728" alt="Screen Shot 2022-08-15 at 7 54 46 AM" src="https://user-images.githubusercontent.com/35935196/184630921-a73a3bea-09b7-477a-867b-e8f914f8fa81.png">

## Define a webhook endpoint for the API Namespace - Web-hook interface
Eg: POST request that must have the following attributes in the JSON payload 

``` javascript
{
  event: 'event_name_string_underscored',
  description: 'human readable description'
}
```
For every web hook triggered External API connection, list its interface under `Webhooks`
<img width="1137" alt="Screen Shot 2022-08-15 at 7 54 31 AM" src="https://user-images.githubusercontent.com/35935196/184630998-8d0a5f8b-d12a-49d0-a22a-e7c1c9bd3a35.png">

## Baked in Authentication (nice to have)
<img width="1159" alt="Screen Shot 2022-08-15 at 7 55 47 AM" src="https://user-images.githubusercontent.com/35935196/184631988-defa3fb6-b948-4100-957a-2f349f98eaeb.png">
<img width="1118" alt="Screen Shot 2022-08-15 at 7 55 34 AM" src="https://user-images.githubusercontent.com/35935196/184631990-57bc03ba-89df-4c35-bb33-b766c1f6e66a.png">

Rename API Clients => API Keys. Decouple API Clients (API Keys) from the REST interface and the API Namespace. Allow API Clients (API Keys) to be issued and be associated with many API Namespaces

Pralish:
@donrestarone , What is event in this payload?

```
{
  event: 'event_name_string_underscored',
  description: 'human readable description'
}
```

donrestarone:
@Pralish the event name is a unique identifier in case an endpoint needs to handle different scenarios 


Eg: event for failed transactions vs successful ones 

See Stripe webhook docs: https://stripe.com/docs/webhooks

We are looking to support a service like that (with web hook authentication and events)

PR #1241, opened 2022-11-23
[feature] Web hook drive strategy + Decouple api clients

donrestarone:
Addresses: https://github.com/restarone/violet_rails/issues/682

Api Keys:

https://user-images.githubusercontent.com/50227291/196992417-ab91ed2e-1514-4abf-9954-d44123542583.mov

Webhook:

https://user-images.githubusercontent.com/50227291/196992487-4c3c9593-4cb6-4b8c-8f7b-c6e22d9385a2.mov

**The webhook's payload can be accessed as `parameters[:request]['body']` in external api client.**

Eg: https://github.com/restarone/violet_rails/pull/1167/files#diff-1aa9cadc4b4c76bf7ec3d7c0cf8d0da66be3ba87336bad4d32241c99d53aa6beR603-R623

```
    class ExternalApiModelExample
      def initialize(parameters)
        @external_api_client = parameters[:external_api_client]
        @request = parameters[:request]
      end
  
      def start
        if @request['body']['type'] == 'customer.created'
          @external_api_client.api_namespace.api_resources.create(
            properties: {
              request_body: @request["body"]	
            }
          )
        end
      end
```

Custom webhook verification method:

https://user-images.githubusercontent.com/50227291/197921958-a06b85a8-7c7a-4e3b-afb9-889fe5c6e110.mov


donrestarone:
@Pralish please see below for my feedback: 

1. can you add the HTTP method in the generated documentation? I believe it should be POST
Screen Shot 2022-11-22 at 9 28 03 PM

2. When stripe is selected, the label should be Webhook signing secret

Screen Shot 2022-11-22 at 9 31 48 PM

3. Its not clear in the inline documentation what `verification_method` means. What exactly is it? I thought that verification method was the custom method definition that would be entered in the form 
Screen Shot 2022-11-22 at 9 31 37 PM

donrestarone:
@kripananda-yadav to coordinate with @Pralish to create a permanent test case for this feature. I recommend creating an endpoint that takes a date range and exposes cars created between that range (https://violet-test.net/car-renderer)

Pralish:
@donrestarone , I have addressed your [feedback](https://github.com/restarone/violet_rails/pull/1241#issuecomment-1324472982).

Let me know what you think?

Screenshot 2022-11-25 at 12 49 00 PM
Screenshot 2022-11-25 at 12 49 10 PM
Screenshot 2022-11-25 at 12 49 24 PM

donrestarone:
@Pralish could you fix the merge conflicts here when you get a chance?

donrestarone:
@Pralish there is a test failure related to API access by category (CC @alis-khadka ) -- can you address it so @kripananda-yadav can start UAT?

donrestarone:
> @donrestarone , I have addressed your [feedback](https://github.com/restarone/violet_rails/pull/1241#issuecomment-1324472982).
> 
> Let me know what you think?
> 
> Screenshot 2022-11-25 at 12 49 00 PM
> 
> Screenshot 2022-11-25 at 12 49 10 PM
> 
> Screenshot 2022-11-25 at 12 49 24 PM

Inline docs look good! can you write up a GitHub issue that documents how to setup a web hook connection class? After release we can use that GitHub issue to update the docs: https://github.com/restarone/violet_rails/wiki/API:-Entities,-Form-Rendering,-Interfaces-and-Actions#automation-connecting-and-extracting-data-from-external-systemsapis

kripananda-yadav:
### Acceptance Criteria:(In Progress)

When user creates   a endpoint or API External client and place drive strategy as webhook then it should be displayed under the web hook section of the Namespace.
User should not be able access the endpoint if it’s not enabled.
User should only access the endpoint if enabled.
User needs authorization key if “Required webhook verification ” is enabled.
Route should not exist if drive strategy id not webhook
When user tries to access the endpoint should return error response if webhook verification required and webhook verfication failed

donrestarone:
@kripananda-yadav please save your UAT code so you could create a permanent external API connection for acceptance testing in the main testnet (violet-test.net) for future

alis-khadka:
> @Pralish there is a test failure related to API access by category (CC @alis-khadka ) -- can you address it so @kripananda-yadav can start UAT?

@donrestarone Shall I add the following access for `api_keys` controller?
- read access for api_keys only
- full access for api_keys only
- delete access for api_keys only

donrestarone:
@alis-khadka @kripananda-yadav is this ready for UAT?

alis-khadka:
> @alis-khadka @kripananda-yadav is this ready for UAT? 

@donrestarone some code refactoring part is remaining. I will update in this thread after it is done. It should be done by today.

alis-khadka:
@Pralish It is ready for to take on as you said some change needs to be done on the ApiClients.

kripananda-yadav:
@alis-khadka When user has only the delete access for the API key, User is able to click the show button as of now there is no new data that is exposed but this case might raise problem.

https://user-images.githubusercontent.com/70195786/209466973-21744dc0-5988-47e6-b6ed-00f4664b0f7d.mov

alis-khadka:
> @alis-khadka When user has only the delete access for the API key, User is able to click the show button as of now there is no new data that is exposed but this case might raise problem.
> 
>  Screen.Recording.2022-12-25.at.5.40.24.PM.mov

It is a reasonable point but I think there should be a `delete` button in the show view of ApiKey.

@donrestarone What's you view on this? Shall I add a `delete` button in the show-view of Apikey or restrict the show-view if the user has only `delete` access?

donrestarone:
@alis-khadka you need to do both. Restrict the show view + add a delete button if access is only for delete

donrestarone:
@kripananda-yadav could you expand on your acceptance criteria for the stripe and custom web hook verification

> Authentication using Both Stripe and custom
User needs authorization key if “Required webhook verification ” is enabled.
Route should not exist if drive strategy id is not webhook
When user tries to access the endpoint should return error response if webhook verification required and webhook verification failed
User needs authorization key if “Required webhook verification ” is enabled.
When user is able to access the endpoint the class method definition should be executed and result must be reflected as well in the expected portions.


you haven't covered: 

1. returning a boolean by executing ruby in a custom web hook verification (access grant vs access denied responses)
2. when web hook secret is not provided in a custom web hook verification
3. using an API key for the web hook secret


@Pralish the API clients is leading to the Connections page, is that expected?

Screen Shot 2022-12-25 at 10 11 56 AM

also, please use all caps for API in `API Keys`

Screen Shot 2022-12-25 at 10 17 11 AM
Screen Shot 2022-12-25 at 10 17 04 AM


Screen Shot 2022-12-25 at 10 17 18 AM

alis-khadka:
> @Pralish the API clients is leading to the Connections page, is that expected?
> 
> Screen Shot 2022-12-25 at 10 11 56 AM
> 
> also, please use all caps for API in `API Keys`
> 
> Screen Shot 2022-12-25 at 10 17 11 AM
> 
> Screen Shot 2022-12-25 at 10 17 04 AM
> 
> Screen Shot 2022-12-25 at 10 17 18 AM

Actually, the entity's model name is **ExternalApiClient**. On other pages, we are using the text **'External API Connection'**. So, shall I change the text from **'create new External API Client'** to **'create new External API Connection'** ?

kripananda-yadav:
@Pralish 
- Model Definition is not executed when user access the endpoint which  is create in the External API client gets success as response.
- The response code is 200 even if the authentication fails while accessing the Namespaces when user has enabled authentication and had entered invalid authentication token.
Screenshot 2022-12-27 at 5 26 49 PM

For the API created using External API client:
- While accessing api using invalid authentication token response code is 400.
For authentication failure the response code should be 401.

donrestarone:
@Pralish the Connection can return a response to the API caller right? Eg:

``` ruby
    class ExternalApiModelExample
      def initialize(parameters)
        @external_api_client = parameters[:external_api_client]
        @request = parameters[:request]
      end
  
      def start
        if @request['body']['type'] == 'customer.created'
          created_object = @external_api_client.api_namespace.api_resources.create(
            properties: {
              request_body: @request["body"]	
            }
          )
          render json: created_object
        end
      end
```

Pralish:
> @Pralish the Connection can return a response to the API caller right? Eg:
> 
> ```ruby
>     class ExternalApiModelExample
>       def initialize(parameters)
>         @external_api_client = parameters[:external_api_client]
>         @request = parameters[:request]
>       end
>   
>       def start
>         if @request['body']['type'] == 'customer.created'
>           created_object = @external_api_client.api_namespace.api_resources.create(
>             properties: {
>               request_body: @request["body"]	
>             }
>           )
>           render json: created_object
>         end
>       end
> ```

@donrestarone , It cannot. Since  webhooks response should be fast and short, it only returns a status code and error message incase of error.

https://github.com/restarone/violet_rails/pull/1241/files#diff-65660574e19af12662666fdd223f31bc76086d728d8e56c6fa916591bafacc21R14-R19

Shall I update the code to return custom response?

Pralish:
@donrestarone , The ability to render customized response on webhook has been added.

donrestarone:
@Pralish great. could you please update, 

1. The short docs here (https://github.com/restarone/violet_rails/pull/1241#issue-1460830049)
2. The generated in-line doc on the app

to reflect sending a JSON response

donrestarone:
@kripananda-yadav please add the JSON response criteria to your acceptance criteria, I suggest the following: 

1. in cases without explicit return value, web hook returns 200 response for successful requests
2. if an explicit return value is defined, it returns a JSON response

donrestarone:
@Pralish I did a UAT pass and found the following, 

1. Driving a Connection via web hook is broken:

Screen Shot 2023-01-07 at 11 57 05 AM

when the same class definition is called on-demand, it works:

Screen Shot 2023-01-07 at 11 57 12 AM


UX bugs:

1. Drive interval is not needed when driven by web hook: 

Screen Shot 2023-01-07 at 11 47 54 AM

2. Fix name: 

Screen Shot 2023-01-07 at 11 49 40 AM

3. remove nav tab, link directly to page:

Screen Shot 2023-01-07 at 11 50 37 AM

4. Need breadcrumb on (https://review-1241.violet-test.net/api_namespaces/25/external_api_clients/4): 

Screen Shot 2023-01-07 at 11 51 27 AM

Screen Shot 2023-01-07 at 12 02 01 PM

donrestarone:
@Pralish this is coming together well, some clarifications: 

1. please use the following boilerplate class for the web hook connection:

``` ruby

class WebhookDrivenConnection
  def initialize(parameters)
    @external_api_client = parameters[:external_api_client]

    # rails request object accessable for webhook, https://api.rubyonrails.org/classes/ActionDispatch/Request.html
    @payload = parameters[:request]&.request_parameters
  end

  def start
    object = @external_api_client.api_namespace.api_resources.create(
      properties: {
        request_body: @payload
      }
    )
    # render the object as the response
    render json: { result: object }
  end
end

WebhookDrivenConnection
```

2. Ensure that the Connection button is where it always was (we simply want to remove the extra step in opening a nav drawer). It should link directly just like **Actions** does

Screen Shot 2023-01-09 at 12 18 23 PM

app/models/external_api_client.rb
```diff
@@ -41,7 +47,11 @@ class ExternalApiClient < ApplicationRecord
   validates :drive_every, presence: true, if: -> { drive_strategy == ExternalApiClient::DRIVE_STRATEGIES[:cron] }
 
   validates :drive_every, inclusion: { in: ExternalApiClient::DRIVE_INTERVALS.keys.map(&:to_s) }, allow_blank: true, allow_nil: true
-  validates :model_definition, safe_executable: true
+  validates :model_definition, safe_executable: { skip_keywords: ['render'] }
```
donrestarone, line 50: @Pralish could you make the skippable keywords a const (also, how is `render` a blocked world)

donrestarone:
@Pralish I tested the custom validator, and it works well: 

1. Use ahoy visit: 

Screen Shot 2023-01-09 at 3 36 35 PM
2. If ahoy visitor token passed is valid run web hook:

app/controllers/api/base_controller.rb
```diff
@@ -4,7 +4,7 @@ class Api::BaseController < ActionController::API
   def authenticate_request
     if @api_namespace.requires_authentication
       unless validate_bearer_token
-        render json: { status: 'unauthorized', code: 401 }
+        render json: { status: 'unauthorized', code: 401 }, status: 401
```
alis-khadka, line 7: It would be better to use `:unauthorized` instead of `401`;
`render json: { status: 'unauthorized', code: 401 }, status: :unauthorized`

app/controllers/api/external_api_clients_controller.rb
```diff
@@ -0,0 +1,40 @@
...
+    def webhook
+      # should only exist if drive strategy is webhook
+      raise ActionController::RoutingError.new('Not Found') unless @external_api_client.drive_strategy == ExternalApiClient::DRIVE_STRATEGIES[:webhook]
+
+      return render json: { message: 'Webhook not enabled', success: false }, status: 400 unless @external_api_client.enabled
+
+      if @external_api_client.webhook_verification_method.present?
+        verified, message = Webhook::Verification.new(request, @external_api_client.webhook_verification_method).call
+
+        return render json: { message: message, success: false }, status: 401 unless verified
```
alis-khadka, line 15: Same here. We can use `:unauthorized` symbol.

db/migrate/20221221073259_sanitize_user_accessibility_in_users.rb
```diff
@@ -0,0 +1,21 @@
+class SanitizeUserAccessibilityInUsers < ActiveRecord::Migration[6.1]
+  def up
+    User.all.each do |user|
```
donrestarone, line 3: @Pralish @alis-khadka what's happening here? Looks like the granular permission scheme for API is being changed to accommodate this feature

donrestarone:
I tested the migrations and noticed that the checkbox logic for the permissions should be changed, @Pralish please see below: 

Screen Shot 2023-01-12 at 7 38 25 PM
Screen Shot 2023-01-12 at 7 36 56 PM

db/migrate/20221221073259_sanitize_user_accessibility_in_users.rb
alis-khadka: Before introducing `ApiKey`, all other API entities would fall under an `ApiNamespace`, i.e.: with all other API entities, ApiNamespace had `has many` relation. Due to this, the value of column `api_accessibility` was tailored considering `ApiNamespace`.
```
{ all_namespaces: {full_access: 'true'}}
OR
{namespaces_by_category: {category_name: {full_access: 'true'}}}
```

With the introduction of `ApiKey`, it has `many-to-many` relation with `ApiNamespace` due to which access related to `ApiKey` would not fall under a single `ApiNamespace`. Since,  the value of column `api_accessibility` was tailored considering `ApiNamespace` only, it would be better to organize `api_accessibility` under 2 categories of access; 1) api_namespace  2) api_key. 
```
{
api_namespaces: { all_namespaces: {full_access: 'true'}},
api_keys: {full_access: 'true'}
}
```

This migration file is basically moving the current value of `api_accessibility` under  `api_namespace` key.

donrestarone:
@Pralish @alis-khadka ⚠️ dangerous migration

I created a connection to mimic existing data, and it was overwritten by the default value. This PR cannot be shipped until the migration is fixed and the following data corruption issue is resolved:

Screen Shot 2023-01-13 at 10 53 20 AM

PR #1216, opened 2022-11-11
[feature] 2FA + ability to enforce 2FA in Subdomain

donrestarone:
Addresses: #1081 

Demo: 

https://user-images.githubusercontent.com/55496123/195891042-e9508cb0-0d07-4b1c-9e1c-f0fcfc2659a4.mov

Co-authored-by: Er. Ajay Shrestha <55496123+ajaystha42@users.noreply.github.com>
Co-authored-by: Ajay Shrestha <er.ajaystha@gmail.com>
Co-authored-by: Prashant Khadka <prashant.khadkda052@gmail.com>

PR #1244, opened 2022-11-25
[feature] 2FA + ability to enforce 2FA in Subdomain (#1216)

ajaystha42:


donrestarone:
UAT pass underway: 

1. User is able to reset password when 2FA enabled
2. User is able to reset password when 2FA not enabled
3.  Running the script enables it for all subdomains

Issue #1269, opened 2022-12-04
undefined local variable or method `cookies' for Api::ResourceController


donrestarone:
``` txt 
A NameError occurred in resource#index:

  undefined local variable or method `cookies' for #<Api::ResourceController:0x0055ab9f023eb0>
Did you mean?  cookies_url
  app/controllers/concerns/ahoy_controller_patch.rb:13:in `tracking_enabled?'


-------------------------------
Request:
-------------------------------

  * URL        : https://nikean.org/api/1/story_platform-select-reason_tags
  * HTTP Method: GET
  * IP address : 45.248.151.13
  * Parameters : {"controller"=>"api/resource", "action"=>"index", "version"=>"1", "api_namespace"=>"story_platform-select-reason_tags"}
  * Timestamp  : 2022-12-04 05:36:37 UTC
  * Server : ip-172-31-14-250
    * Rails root : /var/www/violet/releases/20221204023959
  * Process: 907719

-------------------------------
Session:
-------------------------------

  * session id: [FILTERED]
  * data: {"session_id"=>"b9866d74031cb4cf0c4d6d3855776ac6",
   "user_return_to"=>
    "/search?properties={%22tags%22:{%22value%22:[%22psilocybin%22],%20%22option%22:%22PARTIAL%22}}",
   "warden.user.user.key"=>[[19], "$2a$12$LGRk/r9vwnmf/BumeTdVHu"],
   "warden.user.user.session"=>{"last_request_at"=>1670132194},
   "_csrf_token"=>"49HzrmNOMTNlO2AAy9ZggZ8VvxVgdsrIkz4kB8u-nvk=",
   "site_id"=>1}

-------------------------------
Environment:
-------------------------------

  * GATEWAY_INTERFACE                                       : CGI/1.2
    * HTTP_ACCEPT                                             : */*
    * HTTP_ACCEPT_ENCODING                                    : gzip, deflate, br
    * HTTP_ACCEPT_LANGUAGE                                    : en-US,en;q=0.9
    * HTTP_CACHE_CONTROL                                      : no-cache
    * HTTP_CONNECTION                                         : close
[...]

PR #1272, opened 2022-12-04
[bug fix]: undefined `cookies' for Api::ResourceController

alis-khadka:
Addresses: https://github.com/restarone/violet_rails/issues/1269

### DIFF
--- a/test/controllers/api/resource_controller_test.rb
+++ b/test/controllers/api/resource_controller_test.rb
@@ -2,6 +2,7 @@ require "test_helper"
 
 class Api::ResourceControllerTest < ActionDispatch::IntegrationTest
   setup do
+    @subdomain = Subdomain.current
     @api_namespace = api_namespaces(:one)
     @users_namespace = api_namespaces(:users)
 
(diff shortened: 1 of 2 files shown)


test/controllers/api/resource_controller_test.rb
```diff
@@ -27,8 +28,27 @@ class Api::ResourceControllerTest < ActionDispatch::IntegrationTest
       })
   end
 
-  test 'describe resource name and version: get #index as json' do
-    get api_url(version: @api_namespace.version, api_namespace: @api_namespace.slug), as: :json
+  test 'describe resource name and version: get #index as json with no ahoy-visit being tracked when tracking is disabled' do
```
donrestarone, line 31: @alis-khadka can you add a test for the index endpoint as well? This is what is erroring on `nikean.org`
``` txt
Request URL: https://nikean.org/api/1/story_platform-select-reason_tags
Request Method: GET
Status Code: 500 Internal Server Error
Remote Address: 3.99.30.108:443
Referrer Policy: strict-origin-when-cross-origin
```
Screen Shot 2022-12-04 at 11 21 09 AM

alis-khadka: Screen Shot 2022-12-04 at 10 28 13 PM
@donrestarone  Isn't this the index endpoint? This test case is hitting the index endpoint.

donrestarone: @alis-khadka ah okay, as long as your tests are raising the same error without your patch in place we are good

Issue #1315, opened 2022-12-20
uncategorized access should allow access to API Namespace creation


Mamsheikh:
![contractors-Admin](https://user-images.githubusercontent.com/36868017/208691215-dab20857-9364-4a53-8970-1d989ff24ff0.png)

donrestarone:
If the user has access to uncategorized namespaces, they should be able to create new namespaces from scratch

<img width="1728" alt="Screen Shot 2022-12-20 at 6 45 08 AM" src="https://user-images.githubusercontent.com/35935196/208758183-60470511-a862-4669-8602-33e5c2da6316.png">

alis-khadka:
@donrestarone Should we hide **categories** part from the form when a user with **uncategorized** access is creating an API Namespace?
<img width="816" alt="Screen Shot 2022-12-21 at 9 10 08 PM" src="https://user-images.githubusercontent.com/25191509/208941641-1ff64cfe-3e60-437c-a377-c44210a3daca.png">

PR #1316, opened 2022-12-22
[feature] Uncategorized access should allow API Namespace creation

alis-khadka:
Addresses: https://github.com/restarone/violet_rails/issues/1315

**Demo**

https://user-images.githubusercontent.com/25191509/209044452-9db5a0f7-8037-4561-bdf9-0f06c99a6508.mov

PR #1319, opened 2022-12-23
[feature] Uncategorized access should allow API Namespace creation

donrestarone:
Addresses: https://github.com/restarone/violet_rails/issues/1315

**Demo**

https://user-images.githubusercontent.com/25191509/209044452-9db5a0f7-8037-4561-bdf9-0f06c99a6508.mov


donrestarone:
passing UAT run: https://github.com/restarone/violet_rails/pull/1316

Issue #1305, opened 2022-12-17
add permission can access forum


donrestarone:
add permission can_access_forum, backfill `true` so that existing forum users are not impacted. Default value moving forward should be `false`

PR #1325, opened 2022-12-28
[feature] permissioned forum access

alis-khadka:
Addresses: https://github.com/restarone/violet_rails/issues/1305

**Demo**

https://user-images.githubusercontent.com/25191509/210078509-2b229519-0028-4b4a-8c59-7f5c9d2d137f.mov


donrestarone:
Acceptance criteria: 

1. When forum is public, visitors should have read access  ⌛ 
2. When forum is public, users without access permission should have read access ⌛ 
3. When forum is public, users with access permission should have write access ⌛ 
4. When forum is private, users without access permission should not have read access ⌛ 
5. When forum is private, users with access permission should have read/write access ⌛ 
6. Users with moderator permission should have ability to access forum categories ⌛ 
7. Users with moderator permission should have ability to modify any post/thread ⌛

donrestarone:
@alis-khadka to fix the PR so it adheres to [this](https://github.com/restarone/violet_rails/pull/1325#issuecomment-1368517983)

alis-khadka:
@donrestarone 

1. When forum is `public`, should we restrict the visitors and users without access permission from creating forum thread/post ?
2. Should we allow only users with moderator permission, to create new `forum categories`?
3. And, users with moderator permission will bypass the access permission, right?

donrestarone:
@alis-khadka `can manage forum supersedes `can access forum` so when `can manage forum` is checked, we should auto check the `can access forum` checkbox to indicate it to the user

Screen Shot 2023-01-05 at 10 09 18 AM

Issue #1292, opened 2022-12-11
entity specific analytics


donrestarone:
**Is your feature request related to a problem? Please describe.**

original thread: https://nikean.org/forum/threads/custom-analytics-dashboard

**Describe the solution you'd like**

see prototype: https://xd.adobe.com/view/efad6f4d-bd44-4f7d-85b0-3679e3797dc0-bafe/

### overview 

![Screenshot 2022-12-11 at 11-13-09 Nikean Dashboard](https://user-images.githubusercontent.com/35935196/206915988-e1617818-5a6b-4bf9-9ec8-cb9c3b0f76af.png)

###  add JSON scheme to track meta data in Ahoy event
Create a JSON spec for adding the relevant metadata to the properties object in the `ahoy_events` table:

 
<img width="965" alt="Screen Shot 2022-12-19 at 8 03 51 AM" src="https://user-images.githubusercontent.com/35935196/208432518-1b0581e3-c69d-49d2-96c7-3c7f6d249138.png">

The events are currently used in the present Visitor Analytics in a list view:

<img width="1147" alt="Screen Shot 2022-12-11 at 11 23 38 AM" src="https://user-images.githubusercontent.com/35935196/206915635-31993f64-a097-4f5d-97e6-d15f6f6b5120.png">

### Add Analytics V2 screen

<img width="1728" alt="Screen Shot 2022-12-19 at 8 02 09 AM" src="https://user-images.githubusercontent.com/35935196/208432323-70e3c1db-9cf3-4ef0-be30-1b17af864c6a.png">

## Client

create a contract that client engineers can uphold when rigging up analytics using Ahoy.JS -- here's how we rig for analytics now: https://github.com/restarone/violet_rails/wiki/%5BFeature%5D-Analytics-&-Reporting#visitor-analytics-on-nikean-website

At the moment, `ahoy.track` events have no common standard. In the example below we are accidentally passing an attribute that is not even used. 

<img width="1728" alt="Screen Shot 2023-01-12 at 8 37 24 PM" src="https://user-images.githubusercontent.com/35935196/212217208-cf823697-1494-4b71-8769-beeba4039d05.png">

donrestarone:
Use a json scheme to express how to track the entity event
``` json

{
analytics: {
  name: 'foo',
  type: 'click'
}

}

```

PR #1389, opened 2023-01-30
[feature] Entity specific analytics

donrestarone:
Addresses: https://github.com/restarone/violet_rails/issues/1292 Screen Shot 2023-01-04 at 8 07 53 AM

Co-authored-by: Pralish Kayastha <50227291+Pralish@users.noreply.github.com>
Co-authored-by: Pralish Kayastha <pralishkayastha@gmail.com>


donrestarone:
⚠️  @Pralish this is failing UAT on staging due to the following: 

1. Clickable tracking not working:
Screen Shot 2023-01-30 at 12 01 24 AM

2.  Viewable tracking not working:

Screen Shot 2023-01-29 at 11 59 38 PM


In both cases I have ensured that I have consented to cookies/tracking

donrestarone:
On first load, tracking reliability issue is solved by upgrading to the newer cookie consent buttons instead of anchor links: 

use this
``` html 
      <button class="btn btn-primary mb-3 cookie-consent-button" data-value="true">Accept All</button>
```

instead of: 

``` html
      <a class="btn btn-accept mb-3" href="/cookies?cookies=true">ACCEPT ALL</a>
```

@Pralish section views are still broken. Could you use violet.restarone.solutions to investigate and patch?

donrestarone:
⚠️ running temporary deployment to nikean org (https://github.com/restarone/violet_rails/settings/environments/447769648/edit) to validate watch time tracking:
https://github.com/restarone/violet_rails/actions/runs/4010531168/jobs/6960144816

donrestarone:
video watch time tracking LGTM on nikean.org: 

Screen Shot 2023-01-30 at 1 42 59 PM

Pralish:
### Determining the visibility threshold for section views 

Threshold indicates at what percentage of the target's visibility the tracking should be executed. 0.75 threshold indicates that the tracking is done when 75% of the section is visible in the screen.

However for larger sections, 75% of the content can never be visible. Therefore we need to adjust the threshold.

A good rule of thumb is if the section is double the screen size, adjust the threshold to slightly less than 0.5 and if the section is 4 times the screen size, adjust threshold to less than  0.25, and so on

REF:

https://stackoverflow.com/questions/66296057/intersectionobserver-does-not-work-on-small-screens-for-long-sections-js/

https://developer.mozilla.org/en-US/docs/Web/API/Intersection_Observer_API#thresholds

Issue #1321, opened 2022-12-24
CMS association not detected when API form is rendered


donrestarone:
Affecting ordinarytrip.com &. contractors.restarone.solutions

<img width="1728" alt="Screen Shot 2022-12-24 at 10 16 15 AM" src="https://user-images.githubusercontent.com/35935196/209442174-0d051f27-132c-44a1-8c9e-766039fc9266.png">
<img width="1728" alt="Screen Shot 2022-12-24 at 10 15 56 AM" src="https://user-images.githubusercontent.com/35935196/209442180-6d72f0fc-49a6-4086-b86b-809f07456012.png">

<img width="1151" alt="Screen Shot 2022-12-24 at 10 15 07 AM" src="https://user-images.githubusercontent.com/35935196/209442208-f740a3ee-16c2-4cab-b51b-8187d34d8dab.png">
<img width="953" alt="Screen Shot 2022-12-24 at 10 14 59 AM" src="https://user-images.githubusercontent.com/35935196/209442211-41afdfe2-8892-4343-994d-9efba9314499.png">

PR #1398, opened 2023-02-05
[Bug Fix]: CMS association not detected when API form is rendered

alis-khadka:
Addresses: https://github.com/restarone/violet_rails/issues/1321

**DEMO**

https://user-images.githubusercontent.com/25191509/216809971-c0b7d411-9b0f-4479-a4ed-263f07cb4429.mov


kripananda-yadav:
Acceptance Criteria: 

- When the "Form rendering snippet" Code is added to any page it should be associated with the namespace of whose the snippet code is added. 
- Gap or any html tag should not block the association of the Namespace with the CMS.

Issue #1375, opened 2023-01-20
Cannot change account email from user settings


Ayon95:
**Describe the bug**
When I open the link to confirm account email change, an "Invalid token error" is shown

**To Reproduce**
1. Go to 'My Settings'
2. Provide an email
3. Click on 'Update'
4. Click on the link in the confirmation email sent by Violet Rails

**Expected behavior**
I should be able to confirm changes after going to the confirmation link

**Screenshots**

Changing account email:
![changing-email-from-admin](https://user-images.githubusercontent.com/73725297/213725200-15985b66-680f-4d05-ad3e-c090e6be256b.png)

Confirmation email sent by Violet Rails:
![link-to-confirm-email-change](https://user-images.githubusercontent.com/73725297/213726803-e871a7ed-27da-4aca-9592-f8e6daf1352c.png)

Invalid Token error:
![invalid-token-error](https://user-images.githubusercontent.com/73725297/213726768-ecd68023-79b3-4e83-9b04-23f62bb438f7.png)

**Desktop**
 - OS: Windows 10/11
 - Browser: Chrome 
 - Version: 109.0.5414.75

donrestarone:
this is affecting nikean.org

PR #1400, opened 2023-02-05
[bug fix] Cannot change account email from user settings

Pralish:
Addresses: https://github.com/restarone/violet_rails/issues/1375

Demo:

https://user-images.githubusercontent.com/50227291/216375346-430563ab-382e-4c21-a1f5-2aca95f6e4b2.mov


donrestarone:
@Pralish I cannot change the email in the review app, marking as failing

PR #1405, opened 2023-02-06
[Bug Fix]: CMS association not detected when API form is rendered

alis-khadka:
Addresses: https://github.com/restarone/violet_rails/issues/1321


alis-khadka:
@donrestarone I have added further test cases. We proceed with this PR.

Issue #1083, opened 2022-09-05
add rake task to purge orphaned/unattached active storage blobs


donrestarone:
https://stackoverflow.com/questions/60194474/rails-5-2-active-storage-how-to-determine-and-ensure-that-there-are-no-floating

donrestarone:
nikean.org has had failed uploads that have left orphaned blobs

PR #1417, opened 2023-02-18
purge unattached active storage blob

Pralish:
Addresses: https://github.com/restarone/violet_rails/issues/1083

DEMO:

https://user-images.githubusercontent.com/50227291/219999900-c683255b-018c-4208-8b33-70b541494815.mov

### DIFF
--- a/config/schedule.rb
+++ b/config/schedule.rb
@@ -41,4 +41,5 @@ end
 
 every 1.day do
   rake "maintenance:clear_discarded_api_actions"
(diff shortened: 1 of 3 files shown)


donrestarone:
Test plan: purge ActiveStorage attachments on the review app by manually running the rake task from Dokku

donrestarone:
verified with: 
``` bash
dokku run review-1417 rake active_storage:purge_unattached
```

Issue #538, opened 2022-05-07
implement rack attack to throttle / ban IP addresses


donrestarone:

donrestarone:
affecting Nikean org
![IMG_1891](https://user-images.githubusercontent.com/35935196/219248408-1b1c3c6d-e839-4ab5-be56-e7c225e7bc87.PNG)

PR #1419, opened 2023-02-20
[feature] Rack attack

Pralish:
Addresses: https://github.com/restarone/violet_rails/issues/538


config/initializers/rack_attack.rb
```diff
@@ -0,0 +1,14 @@
+class Rack::Attack
+  # Allows 120 requests/IP in ~1 minutes
+  #        240 requests/IP in ~8 minutes
+  #        480 requests/IP in ~1 hour
+  #        960 requests/IP in ~8 hours (~2,880 requests/day)
+  (2..5).each do |level|
```
donrestarone, line 6: @Pralish please provide tuning via a single variable that has a reasonable default value if not set

donrestarone:
@Pralish please also investigate why the server is unhealthy after terminating a request from rack-timeout.

The request times out: 
``` ruby
A Rack::Timeout::RequestTimeoutException occurred in content#show:

  Request ran for longer than 15000ms 
  config/initializers/rack_timeout.rb:21:in `call'
```
and the puma Server hangs, and is unable to serve subsequent requests

app/mailers/rack_attack_mailer.rb
```diff
@@ -0,0 +1,11 @@
+class RackAttackMailer < ApplicationMailer
+  def limit_exceeded(user, error_limit_exceeded = false)
+    @user = user
```
donrestarone, line 3: @Pralish you should CC the email to global admins as well

donrestarone:
🧪 test plan

1. set ERROR_PER_MINUTE=1 and REQUEST_PER_MINUTE=2 ⌛ 
2. user should not be able to raise more than 1 error a minute, or make more than 2 requests per minute ⌛

.env.test
```diff
@@ -9,4 +9,7 @@ APP_HOST=lvh.me:5250
 RECAPTCHA_SITE_KEY=6Lc6BAAAAAAAAChqRbQZcn_yyyyyyyyyyyyyyyyy
 RECAPTCHA_SECRET_KEY=6Lc6BAAAAAAAAKN3DRm6VA_xxxxxxxxxxxxxxxxx
 SECRET_KEY_BASE='38c72586473e364229897f24f1892f1dc5565776878aa4d8c6bf051258622bd2e923b926ab59b40f912b661216f764d993e8d6b8bbfbc33026e5c954b6c51f9b'
+REQUEST_PER_MINUTE=5
+ERROR_PER_MINUTE=3
+PERIOD_MULTIPLIER=2
```
donrestarone, line 14: @Pralish could you create a GitHub issue with documentation on how to tune this feature with these 3 variables?

Issue #1374, opened 2023-01-20
Ability to query API resources that match any or multiple words of a search term


Ayon95:
Currently, it is not possible to implement a keyword-based search. For example, if "professional development" is searched, a partial-match query does not return results for "professional" and/or "development". It tries to match "professional development" entirely.

This the query that I've attempted to use to implement a keyword-based search:
https://github.com/restarone/violet_rails/blob/master/test/models/concerns/jsonb_search/query_builder_test.rb#L33

donrestarone:
adding a bit more detail: 

<img width="1490" alt="Screen Shot 2023-01-20 at 8 56 00 AM" src="https://user-images.githubusercontent.com/35935196/213714073-c717ae69-8799-45ee-9648-245a4725ba8c.png">

we want the system to given a query like "Professional Development" and have the ability to search for keywords (in this case, delineated by spaces) like "Professional" and "Development"  and show results that include "Professional" and/or "Development"

PR #1442, opened 2023-03-09
[feature] Ability to query API resources that match any or multiple w…

donrestarone:
…ords of search term

Addresses: https://github.com/restarone/violet_rails/issues/1374

**DEMO**

https://user-images.githubusercontent.com/25191509/214291392-23e3c962-31ce-4df3-b3fa-1a2565c978be.mov


donrestarone:
@alis-khadka could you write some client-side documentation before this is merged? We need to document how to use this query

CC @Ayon95 -- this is ready to use on Nikean.org after Prashant writes the documentation

Issue #1452, opened 2023-03-11
add test coverage to profiling 


donrestarone:
profiling was introduced as part of: https://github.com/restarone/violet_rails/releases/tag/0.9.98

we should add 2 tests to ensure that: 

1. `?pp=*` query param commands are only run when the user is an admin and has access to profiling
2. `?pp=env` query param commands is only accessible to global admins

alis-khadka:
@donrestarone 
> 1. ?pp=* query param commands are only run when the user is an admin and has access to profiling

To which column are we referring for `user is an admin`? Well, we do have `global_admin` column in users table.

donrestarone:
@alis-khadka in the case for `?pp=*` we need to make sure the user `can access admin` and `can access profiling`. 

In the case for viewing the ENV (`?pp=env`) we need to make sure the user has all the above rights + `global_admin`

Issue #1399, opened 2023-02-05
fix N+1 query in Analytics V2


donrestarone:
**Describe the bug**

Visiting Analytics V2 causes N+1 query. On markedrestaurant.com the N+1 causes a timeout when the range is 3 months or above: 

<img width="1728" alt="Screen Shot 2023-02-03 at 7 36 50 PM" src="https://user-images.githubusercontent.com/35935196/216735330-29b53c45-0ab0-4343-943c-f7fa9829899c.png">

donrestarone:
In addition to the N+1, there maybe an issue with Analytics V2 that causes a memory leak ⚠️ (evidence WIP)

alis-khadka:
> In addition to the N+1, there maybe an issue with Analytics V2 that causes a memory leak ⚠️ (evidence WIP)

@donrestarone 
Is there anything remaining to be updated here regarding the "memory leak" part?

donrestarone:
@alis-khadka not really, except that you can observe on `top` / `htop` the memory does not get freed after visiting the analytics page. eg; on a fresh boot the app can be using 600MB RAM, and after a user visits the analytics page it will jump to 1.2GB and stay there

PR #1462, opened 2023-03-14
[fix] optimize analytics V2 further + lockdown profiler

alis-khadka:
Addresses: https://github.com/restarone/violet_rails/issues/1399 and https://github.com/restarone/violet_rails/issues/1452


test/controllers/admin/comfy/v2/dashboard_controller_test.rb
```diff
@@ -60,38 +60,93 @@ class Comfy::Admin::V2::DashboardControllerTest < ActionDispatch::IntegrationTes
...
-    assert_equal [video_watch_event_1_page_2.id], assigns(:video_view_events).pluck(:id)
-
-    assert_equal [page_visit_event_1.id], assigns(:previous_period_page_visit_events).pluck(:id)
-    assert_equal [click_event_1.id], assigns(:previous_period_click_events).pluck(:id)
-    assert_equal [video_watch_event_1.id], assigns(:previous_period_video_view_events).pluck(:id)
+    # Page Visit Events
+    page_visit_data = assigns(:page_visit_events)
+    assert_equal page_visit_data[:events_count], 1
+    assert_equal page_visit_data[:previous_period_events_count], 1
+    assert_equal page_visit_data[:pie_chart_data]['Single time visitor'], 1
```
donrestarone, line 67: @alis-khadka is there a reason to switch to hard coded right hand values?

donrestarone:
@alis-khadka please add the script for seeding a large number of analytics data

test/controllers/admin/comfy/v2/dashboard_controller_test.rb
alis-khadka: Nope, I will update these.

alis-khadka:
> @alis-khadka please add the script for seeding a large number of analytics data

I replicated the ahoy_events data figures to about 450,000 through script in rails console locally which took about 1+ hour to finish.

Should I also update the seeding file to populate ahoy_events close to this figure or go for abit lower figure?

donrestarone:
> > @alis-khadka please add the script for seeding a large number of analytics data
> 
> I replicated the ahoy_events data figures to about 450,000 through script in rails console locally which took about 1+ hour to finish.
> 
> Should I also update the seeding file to populate ahoy_events close to this figure or go for abit lower figure?

Lets use that complete data set, but wrap it in a conditional block gated by an environment variable so analytics seeding is opt-in. eg:

``` bash
docker-compose run --rm -e SEED_ANALYTICS=true solutions_app rails db:seed
```

alis-khadka:
We can seed the event analytics data at mass by using following command.
`docker-compose run -e SEED_ANALYTICS=true --rm solutions_app rails db:seed`

alis-khadka:
@Pralish 
I seed the database using the following command and the error is still there.
`docker-compose run -e SEED_ANALYTICS=true --rm solutions_app rails db:seed`

test/controllers/admin/comfy/v2/dashboard_controller_test.rb
```diff
@@ -155,4 +246,44 @@ class Comfy::Admin::V2::DashboardControllerTest < ActionDispatch::IntegrationTes
 
     assert_response :success
   end
+
+  # Rack Mini-Profiler test cases - START
```
donrestarone, line 250: @alis-khadka could you move these mini profiler tests to its own test file?

donrestarone: also you need to add a test to verify that the `pp=env` page is only accessible by `global admin`'s

db/migrate/20230321055733_add_index_to_ahoy_events_name_and_time.rb
```diff
@@ -0,0 +1,6 @@
+class AddIndexToAhoyEventsNameAndTime < ActiveRecord::Migration[6.1]
+  def change
+    add_index :ahoy_events, :name
+    add_index :ahoy_events, :time
```
donrestarone, line 4: this looks like a long running migration, we should get @Pralish 's `strong_migrations` changes into master so we could calculate the risk of this migration

test/helpers/dashboard_helper_test.rb
```diff
@@ -89,65 +41,7 @@ class DashboardHelperTest < ActionView::TestCase
...
-  end
-
-  test 'avg_view_percentange' do
-    mock_video_view_events
-    assert_equal 50.0, avg_view_percentage(Ahoy::Event.where(id: [@video_watch_event_1, @video_watch_event_2, @video_watch_event_3].map(&:id)))
-  end
-
-  private 
-  
-  def mock_video_view_events
```
donrestarone, line 119: @alis-khadka have these tests been replaced?

donrestarone:
# test plan 🧪 

1. admins with privilege to access profiling should be able to access `pp=help` ⌛ 
2.  admins without privilege to access profiling should not be able to access `pp=help` ⌛ 
3.  global admin should be able to access `pp=env` ⌛ 
4. non global admin should not be able to access `pp=env` ⌛ 
5. Analytics should track video views ⌛ 
6. There is little to no drift between the analytics v2 statistics in violet-test.net and the review app ⌛

test/controllers/admin/comfy/v2/dashboard_controller_test.rb
alis-khadka: > could you move these mini profiler tests to its own test file?
Sure, I will separate these tests into a separate one.

> also you need to add a test to verify that the pp=env page is only accessible by global admin's
Yup, It's underway. The PR is still in progress.

test/helpers/dashboard_helper_test.rb
alis-khadka: Nope.

Some calculations like total_watch_time, total_views, avg_view_duration, avg_view_percentage were previously done with helper methods. Now, they have been moved to model methods since the calculations involve working on the table itself.

These tests were moved to model level tests with the following commit:
https://github.com/restarone/violet_rails/pull/1462/commits/2021e6a483ed1d46098eb22ee220b9035729020b

donrestarone:
we can verify if `strong_migrations` are working when https://github.com/restarone/violet_rails/pull/1514 is merged. I expect to see a CI failure due to the dangerous migration on this branch, if not we should double check if strong_migrations are working in CI

alis-khadka:
> @Pralish I seeded the database using the following command and the error is still there. `docker-compose run -e SEED_ANALYTICS=true --rm solutions_app rails db:seed`

@Pralish 
I gave some time into it but was not able to fix it. Let this have a fresh look from your end.

donrestarone:
@Pralish can you confirm strong_migrations is running? We expect a failure but didnt get it

Pralish:
> @Pralish can you confirm strong_migrations is running? We expect a failure but didnt get it

The issue was that the migration file was older than strong_migration integration. The issue is fixed and migration is failing as expected

Pralish:
> @Pralish I gave some time into it but was not able to fix it. Let this have a fresh look from your end.

@alis-khadka , This is now fixed.

donrestarone (dismissed):
Code LGTM

donrestarone:
## Profiling Results 📈 🧪 


### Slight improvements to user experience

When analysis going back 1 year is shown, there is a noticeable performance improvement:

Screen Shot 2023-04-08 at 11 03 31 AM

### Less memory & objects used
When a 1 year analysis is shown, less memory and objects are allocated and retained: 

Screen Shot 2023-04-08 at 11 04 09 AM

### Garbage collector is running consistently 
on a per request basis, we observe that the garbage collector runs before the request is served. Indicating that used memory has been drained and freed to be used for other requests. 

Screen Shot 2023-04-08 at 11 06 48 AM

donrestarone:
failing UAT: https://github.com/restarone/violet_rails/pull/1520#issuecomment-1500915505

Issue #1436, opened 2023-03-08
ActionView::Template::Error) "undefined method `/' for nil:NilClass


donrestarone:
``` txt
An ActionView::Template::Error occurred in dashboard#dashboard:

  undefined method `/' for nil:NilClass
  app/views/comfy/admin/v2/dashboard/dashboard.html.haml:143


-------------------------------
Request:
-------------------------------

  * URL        : https://nikean.org/v2/dashboard
  * HTTP Method: GET
  * IP address : 74.15.198.4
  * Parameters : {"controller"=>"comfy/admin/v2/dashboard", "action"=>"dashboard"}
  * Timestamp  : 2023-03-08 00:10:24 UTC
  * Server : ip-172-31-14-250
    * Rails root : /var/www/violet/releases/20230226160309
  * Process: 940827

-------------------------------
Session:
-------------------------------

  * session id: [FILTERED]
  * data: {"session_id"=>"709d596d1d43da29102b4f5749c8705f",
   "user_return_to"=>"/v2/dashboard",
   "warden.user.user.key"=>[[1], "$2a$12$WzLq2Z7gALlIsGWDuPTQQO"],
   "warden.user.user.session"=>{"last_request_at"=>1678234223},
   "site_id"=>1,
   "_csrf_token"=>"CKOwRwgQr0TWPf8ptWtfcmqww5A8vtdMJKo5VSILNgM="}

-------------------------------
Environment:
-------------------------------

  * GATEWAY_INTERFACE                                       : CGI/1.2
    * HTTP_ACCEPT                                             : text/html,application/xhtml+xml,application/xml;q=0.9,*/*;q=0.8
    * HTTP_ACCEPT_ENCODING                                    : gzip, deflate, br
    * HTTP_ACCEPT_LANGUAGE                                    : en-CA,en-US;q=0.9,en;q=0.8
    * HTTP_CONNECTION                                         : close
[...]

PR #1464, opened 2023-03-15
[Bug Fix] Undefined method for nil in analytics v2

alis-khadka:
Addresses: https://github.com/restarone/violet_rails/issues/1436


donrestarone:
⚠️  temporarily deploying @alis-khadka 's patch to [nikean.org](https://github.com/restarone/violet_rails/settings/environments/447769648/edit)

Issue #1381, opened 2023-01-26
allow API form attribute ordering with JSON representation  


donrestarone:
**Is your feature request related to a problem? Please describe.**
Currently, we use JSONB for the form field mapping that does not preserve the order of the attributes submitted by the client-side

<img width="1156" alt="Screen Shot 2023-01-26 at 1 45 58 PM" src="https://user-images.githubusercontent.com/35935196/214923240-af540f24-884e-43b7-b67d-00c146269054.png">

**Describe the solution you'd like**
use a JSON column to keep track of the form field mapping and use that to derive the form field order

donrestarone:
affecting nikean.org -- we have written custom scripts for the form attribute ordering

Pralish:
@donrestarone , To change the type of properties from jsonb to json with zero downtime, It will be a 5 steps process as suggested by strong_migration:

####  First deployment

Create a temporary column temp_properties
Write data to both columns: properties and temp_properties


#### Second deployment

Route the properties attribute to read/write to the temp_properties column, with read fallback on properties and ultimately on  _properties column, which will be introduced in the third deployment.

Remove the  before_save callback that populates data to  temp_properties column from properties.


#### Third deployment

Rename properties column to _properties.

Rename temp_properties column to properties. After the third deployment, verify in the production database that _properties column has the exact same values as properties.


#### Fourth deployment

Ignore _properties column with the [ignored_columns method](https://wrapbook.engineering/api.rubyonrails.org/classes/ActiveRecord/ModelSchema/ClassMethods.html#method-i-ignored_columns).


#### Fifth deployment

Remove _properties column.


Reference: https://wrapbook.engineering/5-steps-to-change-a-column-type-with-zero-downtime-deployment-in-rails.html

Pralish:
@donrestarone , Can we remove creating ApiNamespace from migration here :

db/migrate/20220420153229_add_plugin_subdomain_events_to_api.rb

and make it part of the seed file or a rake task or Insert record to database with SQL instead of doing ApiNamespace.create!

Model class isn't guaranteed to still be around in the same form in a future version of the application, and running the migrations from scratch in the future might yield errors.

https://stackoverflow.com/a/2667747/12489083

Facing same issue here:

I added a column in ApiNamespace and added a validation/callback on that column. Now when I rerun the migration from scratch, it gives me error since the column isn't yet created at this point.

Although rails suggest not to change old migration files, we are bound to face this issue again in the future.

[...]

PR #1491, opened 2023-03-24
[Feat] change the order of properties for API Form

Pralish:
Addresses: 
https://github.com/restarone/violet_rails/issues/555
https://github.com/restarone/violet_rails/issues/1381

Demo:

https://user-images.githubusercontent.com/50227291/227443792-23ac3476-2ded-468c-a3e2-5aab6ccf1196.mov

### DIFF
--- a/app/models/api_namespace.rb
+++ b/app/models/api_namespace.rb
@@ -99,6 +99,7 @@ class ApiNamespace < ApplicationRecord
     end
   }
 
+  # https://github.com/rails/rails/issues/17706
   scope :for_category, ->(*categories) {
     if (categories = [categories].flatten.compact).present?
       select("DISTINCT ON (#{table.name}.id) #{table.name}.*")
--- a/app/models/api_namespace.rb
+++ b/app/models/api_namespace.rb
@@ -97,7 +97,15 @@ class ApiNamespace < ApplicationRecord
     else
       self.none
     end
-   }
+  }
+
+  scope :for_category, ->(*categories) {
+    if (categories = [categories].flatten.compact).present?
+      select("DISTINCT ON (#{table.name}.id) #{table.name}.*")
+      .joins(categorizations: :category)
+      .where("comfy_cms_categories.label" => categories)
+    end
+  }
 
   def update_api_form
     if has_form == '1'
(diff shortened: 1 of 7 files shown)


app/models/api_namespace.rb
```diff
@@ -97,7 +97,16 @@ class ApiNamespace < ApplicationRecord
     else
       self.none
     end
-   }
+  }
+
```
donrestarone, line 101: @Pralish is this change necessary?

Pralish: @donrestarone , it's not necessary. The indentation was one space ahead, so fixed it.

db/migrate/20230323050938_change_properties_in_api_namespaces_from_jsonb_to_json.rb
```diff
@@ -0,0 +1,13 @@
+class ChangePropertiesInApiNamespacesFromJsonbToJson < ActiveRecord::Migration[6.1]
+  def up
+    remove_index :api_namespaces, name: "index_api_namespaces_on_properties"
+
+    change_column :api_namespaces, :properties, :json
```
donrestarone, line 5: @Pralish I'm worried this is going to lock up the table for a long time in big apps, causing requests to to hang while the DB changes in the background

Can you add `strong_migrations` https://github.com/ankane/strong_migrations and see what it says?

Pralish: @donrestarone, The change is on api_namespaces table. I don't think we have a lot of api_namespaces (less than 50 per app in my opinion).

Having said that, I used the strong_migrations gem and this was the response:

```
== 20230323050938 ChangePropertiesInApiNamespacesFromJsonbToJson: migrating ===
-- remove_index(:api_namespaces, {:name=>"index_api_namespaces_on_properties"})
   -> 0.0019s
rails aborted!
StandardError: An error has occurred, this and all later migrations canceled:

=== Dangerous operation detected #strong_migrations ===

Changing the type of an existing column blocks reads and writes
while the entire table is rewritten. A safer approach is to:

1. Create a new column
2. Write to both columns
3. Backfill data from the old column to the new column
4. Move reads from the old column to the new column
5. Stop writing to the old column
6. Drop the old column 
```

donrestarone: @Pralish yeah that was my fear. Can we add strong migrations to the code base, ensure it runs in CI and then tackle this issue with their suggestion?

donrestarone:
@Pralish we are pivoting this task into supporting multiple form representations. This way we don't need to do 5 deployments and we get to keep the JSONB structure for entity representation

donrestarone:
@Pralish we are pivoting this task to: https://github.com/restarone/violet_rails/issues/1512 --

Issue #1383, opened 2023-01-27
allow API Namespace to auto clean resources


donrestarone:
**Is your feature request related to a problem? Please describe.**

Data sets for logging get really big:
<img width="1728" alt="Screen Shot 2023-01-27 at 8 48 52 AM" src="https://user-images.githubusercontent.com/35935196/215102891-9d4b761a-b39e-409a-afa5-d847da268872.png">

**Describe the solution you'd like**

define a clean up interval at the API Namespace level under Settings. Allow clean up by records older than x amount of time (1 week, 2 weeks, 1 month etc)

**Describe alternatives you've considered**
We currently have a Vacuum job plugin to clean up API Namespace resources. It doesn't seem to be working as fast as we would like: https://github.com/restarone/violet_rails/blob/b43d050d41895d56ef9fd9a6dad01f4261168190/test/controllers/admin/comfy/external_api_clients_controller_test.rb#L515

PR #1492, opened 2023-03-24
[Feat] Allow API Namespace to auto clean resources

Pralish:
Addresses: https://github.com/restarone/violet_rails/issues/1383

https://user-images.githubusercontent.com/50227291/227561742-94afc461-9731-49e5-a759-06492b8b84fd.mov


app/controllers/comfy/admin/api_namespaces_controller.rb
```diff
@@ -84,21 +85,7 @@ def create
 
   # PATCH/PUT /api_namespaces/1 or /api_namespaces/1.json
   def update
-    respond_to do |format|
```
donrestarone, line 87: @Pralish I appreciate the code DRY'ing

donrestarone:
@Pralish since we are changing the view on the API Namespace #show page, lets take this opportunity to move the tab bar to the top:
Screen Shot 2023-03-24 at 3 14 08 PM
Screen Shot 2023-03-24 at 3 13 57 PM

lib/tasks/maintenance.rake
```diff
@@ -33,4 +33,21 @@ namespace :maintenance do
     end
     p "cleared old discarded api_actions @ #{Time.now}"
   end
+
+  task :purge_old_api_resources => [:environment] do 
+    Subdomain.all_with_public_schema.each do |subdomain|
+      Apartment::Tenant.switch subdomain.name do
+        ApiNamespace.where.not(purge_resources_older_than: ApiNamespace::RESOURCES_PURGE_INTERVAL_MAPPING[:never]).each do |api_namespace|
```
donrestarone, line 40: lets read in batches, there is a huge number of records

https://api.rubyonrails.org/classes/ActiveRecord/Batches.html

donrestarone:
🧪 @Pralish  when the clean up frequency is specified by submitting the form, a clean up should happen (unless they specify never, of course)

donrestarone:
🧪 Test plan

1. when `allow settings` privilege is granted, the user is able to access the clean up frequency ⌛ 
2. when `allow settings` privilege is not granted, the user is not  able to access the clean up frequency ⌛ 
3. when the clean up frequency is specified by submitting the form, a clean up should happen ❌ 
4. clean up is performed on eligible resources only ⌛

Pralish:
> @Pralish since we are changing the view on the API Namespace #show page, lets take this opportunity to move the tab bar to the top: Screen Shot 2023-03-24 at 3 14 08 PM Screen Shot 2023-03-24 at 3 13 57 PM

@donrestarone , Does that mean API resources list will be part of the tabs/tab-pane?

donrestarone:
@Pralish no it won't be part of the tabs pane. 

The goal is to show the tab options and resource list without having the user scroll up and down needlessly.

donrestarone:
@Pralish nice touch with the javascript confirmation before enqueueing the purge job!

LGTM ✅

lib/tasks/maintenance.rake
donrestarone: @Pralish awaiting this change before merging ⏫

donrestarone:
after my comment is addressed about reading in batches when purging, we can merge

lib/tasks/maintenance.rake
Pralish: @donrestarone , This line reads the api_namespaces for a single subdomain. 

Here's the method that will read and destroy api_resources in batches.
https://github.com/restarone/violet_rails/pull/1492/files#diff-03241a9bad7359398fa0da5610d4fd762f027297ee8acb1326c900d1d8fe49f1R397-R399

donrestarone: yes, but we should use the same method for performing the destruction in batches. 

The problem right now is that we delete in 2 different ways. The rake task can overload a system from the background if it reads too much data.

donrestarone: also, we should read in batches here `ApiNamespace.where.not(purge_resources_older_than: ApiNamespace::RESOURCES_PURGE_INTERVAL_MAPPING[:never]).each`

donrestarone: actually @Pralish Im seeing it now-- we don't have to read the namespaces in batches. This is fine ✅

PR #1494, opened 2023-03-25
[feature] Rack attack (#1419)

donrestarone:
Addresses: https://github.com/restarone/violet_rails/issues/538


donrestarone:
Im unable to login after the previous deployment 🔴 reverting

PR #1514, opened 2023-04-04
[feature] install strong_migrations

donrestarone:
* install strong_migrations
* update contributing doc


alis-khadka:
@donrestarone 
I think we should update our dokku-CI to remove locks, if present, in the app while deploying review-app.
Screenshot 2023-04-05 at 8 45 54 AM

Re-running the job will probably fix this. But I have observed this multiple times.

donrestarone:
> @donrestarone I think we should update our dokku-CI to remove locks, if present, in the app while deploying review-app. Screenshot 2023-04-05 at 8 45 54 AM
> 
> Re-running the job will probably fix this. But I have observed this multiple times.

@alis-khadka its fine, failure to deploy due to another deploy lock is a minor inconvenience. I have added a section in the contributing docs on how to fix transient issues like this

PR #1516, opened 2023-04-06
style: reduce bit typo misspelling README

guspan-tanadi:
see Features section.


donrestarone:
Thanks for the changes @guspan-tanadi !

PR #1522, opened 2023-04-09
[fix] optimize analytics V2 further + lockdown profiler

alis-khadka:
Addresses: https://github.com/restarone/violet_rails/issues/1399 and https://github.com/restarone/violet_rails/issues/1452


alis-khadka:
# test plan 🧪 

1. admins with privilege to access profiling should be able to access `pp=help` ⌛ 
2.  admins without privilege to access profiling should not be able to access `pp=help`⌛ 
3.  global admin should be able to access `pp=env` ⌛ 
4. non global admin should not be able to access `pp=env` ⌛ 
5. Analytics should track video views ⌛ 
6. There is little to no drift between the analytics v2 statistics in violet-test.net and the review app ⌛

PR #1541, opened 2023-04-22
[WIP] Change SECRET_KEY_BASE without breaking anything

Pralish:


donrestarone:
secret rotator should kick in if it detects the system is ready to rotate its secret (existence of old secret key base, new secret key base etc). It should happen automatically.

PR #1558, opened 2023-05-21
[bug fix] Webhook race condition

Pralish:


app/controllers/api/external_api_clients_controller.rb
```diff
@@ -15,7 +14,11 @@
         return render json: { message: message, success: false }, status: 401 unless verified
       end
 
-      @model_definition = @external_api_client.evaluated_model_definition
+      if @external_api_client.model_definition_class_defined?
+        @model_definition = ExternalApiClient.const_get(@external_api_client.model_definition_class_name)
```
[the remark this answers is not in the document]
donrestarone: @Pralish is this a false positive?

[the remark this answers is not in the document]
Pralish: @donrestarone , that's correct. model_definition_class_name is not a model attribute. It's a method

donrestarone:
# UAT 🧪 

1. once merged to RC make sure the shop works https://violet-test.net/custom-shop - checkout session 
2. and we don't get spammed when checkout session / payment is made

test/models/external_api_client_test.rb
```diff
@@ -170,4 +170,76 @@ class ExternalApiClientTest < ActiveSupport::TestCase
...
+      ExternalApiModelExampleNew
+    "
+
+    external_api_client.update(model_definition: model_definition)
+
+    refute ExternalApiClient.const_defined?(:ExternalApiModelExample)
+    assert ExternalApiClient.const_defined?(:ExternalApiModelExampleNew)
+  end
+
+  test 'should allow save if model defination has syntax error' do
```
donrestarone, line 226: @Pralish what is this test case exercising?

Pralish: @donrestarone , Sorry about that. It's a typo. It should have been 'should not allow save if model defination has syntax error'. Fixed it.

ExternalApiClient won't be valid if model_definition has syntax error. Previously we were able to save even when there were syntax errors. And we would only find out once we ran the plugin. 

In this case, the model definition is missing a 'end'. That's why it won't be saved.

donrestarone: @Pralish no worries-- I just merged the geolocation PR, please catch this up-to master and remove the unsafe label if/when CI passes

Issue #1559, opened 2023-05-22
include geolocation in `metadata` hash in cookies/fetch


donrestarone:
https://violet.restarone.solutions/cookies/fetch should return the cookie status + geolocation of the caller based on IP

PR #1560, opened 2023-05-23
[feature]: include geolocation in cookies metadata hash

alis-khadka:
Addresses: https://github.com/restarone/violet_rails/issues/1559

https://github.com/restarone/violet_rails/assets/25191509/f26ef9a8-c95a-4020-91fc-ea4691b3fbe6

### DIFF
--- a/config/initializers/geocoder.rb
+++ b/config/initializers/geocoder.rb
@@ -1,0 +1,4 @@
+Geocoder.configure(
+  # IP address geocoding service (default :ipinfo_io)
+  ip_lookup: :ipapi_com,
+)
--- a/config/initializers/geocoder.rb
+++ b/config/initializers/geocoder.rb
@@ -1,4 +1,6 @@
 Geocoder.configure(
   # IP address geocoding service (default :ipinfo_io)
   ip_lookup: :ipapi_com,
+  # https://github.com/alexreisner/geocoder#caching
+  cache: Redis.new
 )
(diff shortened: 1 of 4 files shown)


config/initializers/geocoder.rb
```diff
@@ -0,0 +1,4 @@
+Geocoder.configure(
+  # IP address geocoding service (default :ipinfo_io)
+  ip_lookup: :ipapi_com,
```
donrestarone, line 3: @alis-khadka are we aware of any rate limits or any scenario in which this will fail (and how do we fail gracefully)?

alis-khadka: @donrestarone 
- The default provider (`ipinfo.io`) provides [1000 requests in a day and 50.000 requests in a month](https://github.com/alexreisner/geocoder/blob/master/README_API_GUIDE.md#ipinfoio-ipinfo_io). And the provider we are using now (`ip-api.com`) [provides 45 request per minutes](https://github.com/alexreisner/geocoder/blob/master/README_API_GUIDE.md#ipinfoio-ipinfo_io). Well they have [not mentioned about the per day or per month limit for free usage](https://ip-api.com/#:~:text=How%20many%20requests%20can%20I,please%20see%20our%20pro%20service.). So, this provider should provide 64,800 requests in a day.

- Regarding the error handling, geocoder gem will rescue any exceptions raised from the internal API calls as a **default behavior**. I have left this behavior pristine as of now.
Below we can observe that the gem only logs the error but does not throw exception by default. In such case, the location details of the IP will be `null`. Here is a [test case for this scenario](https://github.com/restarone/violet_rails/pull/1560/files#diff-f94b528980f0527e80962c4ca705a427001d83a25e4dd299fc7df1626101cdc4R137-R178).

Screenshot 2023-05-26 at 12 38 06 PM

Issue #1597, opened 2023-09-10
implement calendar functionality


donrestarone:
**Is your feature request related to a problem? Please describe.**
currently there is no calendar feature

**Describe the solution you'd like**
1. ability to show a calendar
2. ability to pick working hours
3. book events (unconfirmed, confirmed, maybe, declined)
4. parse `.ics` files in incoming emails and create event
5. ability to send a link to a booking page

Libraries: 

1. https://github.com/icalendar/icalendar - PORO for calendar, event and includes parsing functionality 
2. https://github.com/excid3/simple_calendar - frontend for rendering a calendar

research: 

1. simple_calendar example implementation: https://medium.com/@zachmb22/how-to-display-calendar-events-using-a-rails-gem-6db0b1e74f62
2. using stimulus JS for a calendar implementation: https://dev.to/timnans/rails-calendar-with-tui-and-stimulusjs-3je

PR #1601, opened 2023-09-13
Calendar and event booking

donrestarone:
addresses: https://github.com/restarone/violet_rails/issues/1597

IMG_7264

todo: 

1. include .vcs file for outlook
2. add validations to meeting model

further reading: 

1. icalendar syncing events: https://joshfrankel.me/blog/lemme-pencil-you-in-using-icalendar-and-rails-to-sync-calendar-events/
2. all the options: https://blog.corsego.com/icalendar-ruby
3. publish? https://stackoverflow.com/questions/55927263/what-does-icalendar-publish-method-do
4. dealing with email client quirkyness when displaying RSVP buttons: https://stackoverflow.com/questions/66102584/when-i-add-method-request-to-icalendar-gmail-stops-recognizing-as-event


Gemfile.lock
```diff
@@ -580,7 +583,7 @@ DEPENDENCIES
   will_paginate
 
 RUBY VERSION
-   ruby 3.0.0p0
+   ruby 2.6.6p146
```
donrestarone, line 586: this change should not be included!

app/controllers/meetings_controller.rb
```diff
@@ -0,0 +1,134 @@
+class MeetingsController < Comfy::Admin::Cms::BaseController
+  before_action :set_meeting, only: %i[ show edit update destroy ]
```
donrestarone, line 2: ensure that the user can access email, so this controller is locked down via authorization

Issue #1609, opened 2023-09-18
".tooltip is not a function" in analytics v2


donrestarone:
**Describe the bug**

**To Reproduce**
1. Go to https://markedrestaurant.com/v2/dashboard
2. see date range picker is broken
3. see error in console

``` txt
Error initializing controller

TypeError: e(...).tooltip is not a function
    at t.value (dashboard_controller.js:7:34)
    at new e (stimulus.js:2331:23)
    at e.value (stimulus.js:2622:19)
    at e.value (stimulus.js:2602:26)
    at e.value (stimulus.js:3222:16)
    at e.value (stimulus.js:3088:23)
    at e.value (stimulus.js:1594:23)
    at e.value (stimulus.js:1478:21)
    at stimulus.js:1463:23
    at Array.forEach (<anonymous>)
```

![Screenshot from 2023-09-18 16-32-56](https://github.com/restarone/violet_rails/assets/35935196/b0d8179d-d55b-4fa6-8671-678d7c209f56)

donrestarone:
last known commit when the view was functional: 66a63e4448624bafbadf17cf503e6added58a9dc

PR #1611, opened 2023-09-19
[Bug Fix] "tooltip is not a function" in analytics V2

donrestarone:
Addresses: https://github.com/restarone/violet_rails/issues/1609

### Demo Clip

https://github.com/restarone/violet_rails/assets/25191509/f8368c2b-e072-4b5c-b9e9-bff3ee17c42a

### DIFF
--- a/app/javascript/packs/comfy/admin/cms.js
+++ b/app/javascript/packs/comfy/admin/cms.js
@@ -1,5 +1,6 @@
 import Sortable from 'sortablejs';
 import jQuery from 'jquery';
+import "bootstrap";
 import "controllers"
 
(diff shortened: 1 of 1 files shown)


donrestarone:
@alis-khadka I noticed that the Shortcuts button (to show expansion for web/forum etc) does not work in any other view than Analytics V2. It looks like Bootstrap JS is not loading everywhere in the CMS admin. Could you take a look? 

Screenshot from 2023-09-19 16-09-39

donrestarone:
patch PR: https://github.com/restarone/violet_rails/pull/1612

PR #1663, opened 2024-11-30
[feature] install blazer

donrestarone:
global admins can now access [Blazer](https://github.com/ankane/blazer), for running SQL queries for BI and Data Science purposes: 

Screen Shot 2024-11-30 at 5 21 25 PM

Screen Shot 2024-11-30 at 5 21 51 PM


donrestarone:
note: 

read .yml files with ERB evaluated: 

``` ruby
APP_CONFIG = YAML.load(ERB.new(File.read("#{Rails.root}/config/blazer.yml")).result)
```

PR #1684, opened 2025-10-16
initialize devise mailer using correct schema (#1683)

restarone-solutions:
* initialize devise mailer using correct schema

* read passed in schema

---------


restarone-solutions:
doesn't fix the problem, closing

PR #1696, opened 2025-11-05
install revolvapp, frontend work - WIP

restarone-solutions:
Screenshot 2025-11-05 at 11 06 52 AM


restarone-solutions:
⚠️  investigate revolvapp broken in review app: 
Screenshot 2025-11-06 at 11 31 55 AM

godie:
The problem was with the pack or webpack thing you was adding the styles and javascript and the thing the webpack was doing is separate in 2 files the javascript and css so then you didn't add the css of the generate compile file with the helper function stylesheet_pack_tag but you already have the javascript with javascript_pack_tag and when the precompile assets occurs well create the two files in this case cms-{hash}.js and cms-{hash}.css

PR #1706, opened 2025-11-22
Create Email Templates

restarone-solutions:
✅ CRUD Email Templates 
✅ Inject dynamic content (copy, names, links) via dynamic segments 
✅ Click to copy email HTML
✅ Test send (with or without dynamic content injection)

Screenshot 2025-11-26 at 9 15 01 AM

### DIFF
--- a/app/helpers/messages_helper.rb
+++ b/app/helpers/messages_helper.rb
@@ -25,29 +25,13 @@ module MessagesHelper
     doc.css('title').remove
     doc.css('link').remove
     
-    # Remove HTML comments (like <!--[if mso]>)
-    doc.xpath('//comment()').remove
+    # Get the HTML string
+    result = doc.to_html
     
-    # Clean up: remove empty text nodes and whitespace-only nodes at the start
-    cleaned_children = []
-    found_real_content = false
-    
-    doc.children.each do |node|
-      # Skip whitespace-only text nodes before real content
-      if node.text? && !found_real_content
-        next if node.text.strip.empty?
-        # Skip standalone text nodes before first element (like "My Email")
-        next unless node.text.strip.empty?
-      end
-      
-      # Mark that we found real content when we hit an element
-      found_real_content = true if node.element?
-      
-      cleaned_children << node if found_real_content
-    end
-    
-    # Rebuild the document with cleaned children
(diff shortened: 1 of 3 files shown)


restarone-solutions:
⚠️ outgoing HTML document is broken, 
Screenshot 2025-11-23 at 10 58 44 AM

⚡  however the copied HTML is being rendered as expected-- indicating this is an issue with encoding/character escaping (how the HTML string is handled before its sent from the backend)
Screenshot 2025-11-23 at 11 00 52 AM

godie:
> it would also be good to not send the default footer if a template is being used

Hi @restarone-solutions let me explain if i got it, need to remove the div trix content and then remove the footer from template but if the code template  you are adding the <re-footer> tag ...

godie:
> thinking about this and looking at the API for trix -- perhaps its better to use a separate column for revolv generated html: https://api.rubyonrails.org/v8.1.1/classes/ActionText/RichText.html#method-i-to_plain_text

@restarone-solutions i tried to plain text but i think is not the solution:
Captura de pantalla 2025-11-25 a la(s) 2 05 43 p m

restarone-solutions:
@godie there is still an issue with rendering, see app/views/e_mailer/ship.html.erb 

there is a trix-content div that is wrapping the revolvapp generated HTML. That still needs to be unwrapped which is why I suggested we use a separate column so we don't have to deal with trix default behaviour.

app/helpers/messages_helper.rb
```diff
@@ -1,2 +1,18 @@
 module MessagesHelper
+    # Extract the visible email content from HTML, removing meta tags, style tags, etc.
```
restarone-solutions: @godie like you have this helper for rendering the email inside the Violet Rails system, you'll need a separate rendering strategy in app/views/e_mailer/ship.html.erb to render the revolvapp HTML without trix tags

godie: well the helper function is doing it, i added the helper to the mailer to reuse, can you validate ship.html.erb is sending the rigth format

restarone-solutions:
@godie rendering is good now. I'll report back after a detailed regression test. Preliminary checks are OK. 

Screenshot 2025-11-25 at 5 03 20 PM

PR #1735, opened 2026-01-03
install reverse lookup gem + use it in exception notifier to detect d…

restarone-solutions:
…ilinquent domains


restarone-solutions:
blocked by: https://github.com/restarone/violet_rails/issues/812

